PR #42, opened 2017-09-12
Fix #41 : Use byte array instead of body string in clients

### LINKED ISSUE #41: Sending Bytes as body alters the size of the payload
When using Request and using bytes as the body the server gets a body that is not the original byte array. It is bigger :(

pm489:
fixed. Why isn't this as good as spring.


s4nchez:
LGTM 👍

PR #47, opened 2017-09-15
Can add helpers to the Dust template engine.

npryce:
Apps can pass a list of scripts to run after Dust is initialised, to register helpers.


daviddenton:
I'm wondering if it would be more consistent to use our Uri object instead of the java.net.URL class. Thoughts?

npryce:
I expect that helper scripts will be compiled into the app as resources, so
it'll be easier for API-clients to just get the resource as a URL than to
convert the URL to a http4k Uri (which will then just get turned back into
a URL anyway).

PR #53, opened 2017-10-11
Http4kWebDriver follows redirects and sets cookies

quii:
This fixes a problem where `Http4kWebDriver` does not follow redirects. 

- Moved some code that is useful for this out of the ClientFilters into extension functions
- Ensure cookies get written between each redirected request

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/ext.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/ext.kt
@@ -1,0 +1,23 @@
+package org.http4k.core
+
+private fun Request.ensureValidMethod(): Request =
+    if (method == Method.GET || method == Method.HEAD) this else method(Method.GET)
+
+private fun Request.newLocation(location: String): Uri {
+    val locationUri = Uri.of(location)
+    return if (locationUri.host.isBlank()) {
+        locationUri.authority(uri.authority).scheme(uri.scheme)
+    } else locationUri
+}
+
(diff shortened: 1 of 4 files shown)


quii:
Maybe? The problem with the existing client filter is that it entirely hides all the redirected requests within it's own state. This is problematic because Http4kWebDriver needs to adjust it's own state (the cookies) for every redirect. Trying to make the existing one perhaps take some kind of hook to facilitate that felt a bit messy.

s4nchez:
The CookieFilter has the concept of “storage” to allow greater control of how cookies are managed. Perhaps a customised webdriver cookie storage + standard redirect and cookie filters could do the trick?

quii:
Yes so i tried it a while ago after your suggestion. It still didnt feel right to me, if you outline the requirements for WebDriver:

1. Read any set cookie headers in responses and store them
2. Send cookies in subsequent requests if applicable
3. Get the _final_ URL that we landed on

Number 3 is especially tricky as ideally it would be nice for the return type to include the final response along with the final URL. But we only get back Response so with the current interface it's not possible. 

It means quite a few ad-hocish changes to the existing interface. These are _specific_ requirements for Web Driver only and (to me) doesn't feel appropriate inside the nice and basic follow redirect. I think I prefer a little duplication in this case. 

In respect to the extension methods; I appreciate you want to keep the interface minimal but they seem like pretty reasonable methods to me, especially if you consider a consumer has to explicitly import them anyway.

s4nchez:
We are pairing to make this change.

Issue #13, opened 2017-05-26
Add functions to set response caching headers


s4nchez:
Something like:

```kotlin
Response(OK).body("so so fresh").noCache()
```

And so on...

daviddenton:
Released in v3.7.0

PR #59, opened 2017-10-30
#13 Create extension methods for Response to add caching headers

k0zakinio:
Hey,

Saw you had an Hacktoberfest issue so thought i'd give it a punt.  Not sure if the location/filename of the ResponseCacheExt.kt is where you'd like it to be.  Also not sure if you'd want to allow the chaining conflicting headers (e.g. Response(OK).public().private() etc.) but left that up to the user.

Let me know what you think

Andy

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/ResponseCacheExt.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/ResponseCacheExt.kt
@@ -1,11 +1,29 @@
 package org.http4k.core
 
-fun Response.noCache() = removeCachingHeaderIfSet().header("Cache-Control", "no-cache")
+fun Response.public() = addCachingHeader("public")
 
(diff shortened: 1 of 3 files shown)


daviddenton:
Hey Andy - thanks for the PR - and sorry it's taken so long to reply - this fell of the radar a bit :)

Anyway - whilst the changes look good (with tests :) ) and it's a good starting point, we think the required implementation is a little more complicated than what we've got in this PR - there is actually more complexity to the CacheControl headers than you've covered here - see here for the plethora of options. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

If you've got time and the inclination, I'd be happy to pair with you to extend it in a more comprehensive way. :)

daviddenton:
@k0zakinio if you get a chance to finish off what we paired on then I'll merge it in. :)

daviddenton:
Released in v3.7.0. Thanks for the PR :)

PR #70, opened 2017-11-30
Request/Response equality does not depend on header order

gypsydave5:
I've had a crack at making equality for Requests and Responses less dependent upon the order of headers.

Now header order is unimportant, unless header field names are repeated, in which case headers with the same field name should be declared in the same order in for equality to hold. As per https://tools.ietf.org/html/rfc7230#section-3.2.2 if I'm reading it right 😄 .


daviddenton:
Thanks for the PR - the functionality looks like what we're aiming for. We're going to merge this in due course, but there's the websockets release that I'm working on at the moment so we're going to concentrate on getting that out first (probably this weekend so you won't have to wait long!)

:)

daviddenton:
This is going into v3.2.3 (releasing now) :)

PR #80, opened 2018-01-04
Adding an optional callback when the websockets connects

tom:
Test is a bit noddy but was useful to prove it worked


daviddenton:
Hey - thanks for the PR. To keep this in sync with the inbound websocket API, I've slightly rejigged this to NOT have the explicit onConnect handlers, but to have that passed to the nonBlocking() call. It's stylistically different, but is consistent - so I decided to go with that - the effect is the same.

New API is in v3.8.0 (releasing now)

PR #81, opened 2018-01-05
[Tom] ading headers and timeout to both blocking and nonblocking webs…

tom:
…ocket

Should default to what is defaulted in downstream contractor, except for headers which defaults to null, but I feel an empty map is cleaner.


daviddenton:
Thanks. FYI, I'm slightly changing the interface to take the Headers typealias instead of the Map, and converting the timeout to a Duration. 

Releasing in 3.10.0

Thanks for the PR! 👏 👏

PR #82, opened 2018-01-09
doc - fix first example in core guide

xhanin:
First example uses an `app` val which is not defined.


s4nchez:
Good spot, @xhanin. Thanks!

PR #85, opened 2018-01-16
New module integrating micrometer metrics

kirderf:
New module: http4k-metrics-micrometer
Integrates micrometer.io with http4k and adds the following server and client filters:
RequestTimer - creates timer that measures the request/response time
RequestCounter - creates counter that counts the number of requests

Both are tagged with the request method, response status and a "response id" that by default is the URI template (cleaned) for server and host for client.


kirderf:
Still outstanding: writing module documentation with examples
Not sure we need a CookBook entry, but it could be useful if we want to show a more detailed example where more metrics from micrometer are added, e.g. Jetty metrics, JVM metrics, etc.

PR #86, opened 2018-01-18
New module http4k-client-jetty

kirderf:
New client implementation based on jetty HttpClient. Supports both sync and async clients, with memory and streaming body modes.


daviddenton:
core is good. any chance you can update the docs as well - just look for instances of `http4k-client-apache` and then update accordingly :)

kirderf:
Of course, I forgot to do that in the changes. Will do that now and submit another PR.

PR #89, opened 2018-01-20
Add support for operationId to Swagger

danschultz:
This adds support for adding an `operationId` field to paths for Swagger definitions. This field allows developers to customize the naming of methods when auto-generating API clients in Swagger. For example, if an endpoint `GET /echo/{message}` has an operationId `echoMessage`, the method name in Swagger's client will be `echoMessage` vs `getEchoMessage`


daviddenton:
Looks good. Thanks for the PR :)

daviddenton:
Released in 3.13.0

PR #99, opened 2018-02-27
Updated to handle up to 10 path parameters in a contract route.

scap1784:
I ran into this issue when I was trying to implement an api with HTTP4k and really need the support for at least 6 path parameters. I added 10 in the hopes of being more future proof.


daviddenton:
Thanks! releasing now in 3.18.0.

PR #110, opened 2018-03-19
Fix conversion of timeout to millis for websocket client

anorth:
Symptom: WS clients close immediately with a connection timeout when given a timeout like `30` (interpreted as milliseconds).


daviddenton:
Good catch. Thanks! 😀

daviddenton:
Releasing in 3.21.2

PR #117, opened 2018-04-06
LensFailure: retain some contextual information [fixes #116]

### LINKED ISSUE #116: How to get more specific LensFailure messages?
Situation: I get a `HTTP/1.1 400 body 'body' must be object` response with an empty body after sending a JSON string to an HttpHandler. I think this message is too generic. Ideally, it should be more specific (give more contextual information).
I had to debug to find out the cause of this exception, which was:
```
com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: For input string: "abc"
```

elifarley:


daviddenton:
Following your comments - I've managed to fix #116 in a different way which I think is more generic. - thanks for the PR though. :) :)

elifarley:
Indeed, your solution is more generic, thanks!

PR #119, opened 2018-04-16
[Tom] updating packages to the point where http4k builds on java 10 w…

tom:
…ith as few warnings as possible and fixing a query parameter bug in the websockets upgrade request where query parameters are repeated


daviddenton:
releasing now in 3.24.0. thanks for the PR! 😄

PR #127, opened 2018-05-11
OpenApi example values

skewwhiffy:
For example objects in a schema, this PR will populate the example field of OpenAPI so that end users will see an example request rather than simply `"string"` or `0`.


http4k-contract/src/test/kotlin/org/http4k/util/JsonToJsonSchemaTest.kt
```diff
@@ -19,7 +19,7 @@ class JsonToJsonSchemaTest {
     fun `renders all different types of json value as expected`() {
         val model = json.obj(
             "aString" to json.string("aStringValue"),
-            "aNumber" to json.number(BigDecimal(1.9)),
+            "aNumber" to json.number(BigDecimal("1.9")),
```
skewwhiffy: 1.9 was getting rounded to 1.89999878987 because reasons.

daviddenton:
LGTM! Thanks :)

Issue #118, opened 2018-04-15
Chaos monkey module


daviddenton:
As touched on in the "Server as a Function. In Kotlin" presentation, it would be awesome to have a separate module which introduced bad behaviour to http4k applications.

This could be easily modelled by a set of filters which generated the chaos behaviours in http4k apps.

Inspired by this: https://codecentric.github.io/chaos-monkey-spring-boot/

IgorPerikov:
Hi @daviddenton! I'd like to work on this one. I've started looking for http4k features, and I can't find whether it's possible to inject filter conditionally without touching application code or not? It doesn't feel right, if one will explicitly put his application handler in chain with latency injection filter and enable/disable it by passing java/env var.

daviddenton:
Thanks for the offer of help @IgorPerikov! You're correct - there currently isn't a way to conditionally inject a filter into the stack. Then again, part of this feature idea was to determine exactly what and how we can do to inject "bad behaviour" into an application.

Initial thoughts were that this was going to evolve through several iterations:

1. Initially, identify the set of the behaviours we're like to model (latency/random failure etc)
2. Implement a set of Filters to inject these behaviours in a purely random fashion.
3. Create some kind of more detailed model/DSL for configuring the randomness - not sure if this might vary based upon the specific behaviour. For example, percentage occurance, initial delay, failure interval or activation predicate.
4. Potentially investigate doing injection of this behvaiour at runtime instead of compile time - we're not a big proponent of  classloading magic, so we'd like to avoid diving right into the deep end on this right away.

Th nice thing about the HttpHandler and Filter model is that it works at several layers of abstraction (endpoint/module/entire app) so instrumenting a running version of your http4k application can be done without implementing step 4 above.

IgorPerikov:
@daviddenton Before I wrote my message, I had already written draft of latency injector (with exception throwing filter in mind)
And here's brief description of what I've coded - latency filter implementation which can be constructed in a pretty old-style fashion with companion method which gather some env variables like that:
`CHAOS_LATENCY_INJECTION_ENABLED`
`CHAOS_LATENCY_INJECTION_MIN_DELAY`
`CHAOS_LATENCY_INJECTION_MAX_DELAY`
`CHAOS_LATENCY_INJECTION_RATE`
or explicitly through passing these to constructor and then you simply apply it to route/app etc

[...]

PR #130, opened 2018-05-23
Chaos testing module

IgorPerikov:
According to https://github.com/http4k/http4k/issues/118
That's what I came up with - a draft implementation of latency injection filter

I also added header setting, so one can understand whether the request is slow because of chaos injection or some other reasons

cc @daviddenton


IgorPerikov:
about failed checks:

- 1 introduced from codebeat - duplicate code, it's ok since it's a draft to show what I had in mind
- coverage decreased - same, one of the filters is not covered
- travis docs section failed - I don't understand yet what this about, I guess I'll figure out later on with that

daviddenton:
@IgorPerikov Don't worry about the docs thing - it's a known issue that the build attempts to perform the build docs step on the PR branches. We've tried to fix that a couple of times already but it seems to be elusive! Haven't had a chance to look over your code yet - will try to have a look soon. :)

daviddenton:
Obviously this is still WIP, but a couple of things about how we would envisage this going. Overall you're on the right track - just pointers really - not criticism! :) :

1. Like the idea of SysEnv instead of sysprops - more 12 factor app etc.. :) 
2. We'd want this stuff be a new module - maybe `http4k-testing-chaos`?. Obviously that's a bit of work to add in all the right places, but you can follow the examples for something simple like the webdriver module - just search for occurances of webdriver and go from there. :)
3. It would be nice to separate the algorithm for injecting random behaviour (maybe a `ChaosPolicy` typealias/interface?), away from the Filters injecting that behaviour. This would accomplish 2 things: leaning on delegation allows users to develop their own policies/behaviours (one of which is the system environmental variable option that you've already come up with), and it makes things easier to test (rather than conflating the randomisation and the behavioural effect). So taking this further, you'd probably get 3 overall concepts:
a. The "ChaosPolicy" to define a couple of predicates which determine if a behaviour should be applied for a request or response (this allows you to do it before or after the downstream call). One of the policy implementations (the default) would be to never apply any behaviours.
b. "Behaviours" - these apply the actual behaviour - e.g latency or blowing up etc.
c. A filter which coordinates #a and #b - so it checks to see if a behaviour should be activated for a request (or a response).

4. With regards to the environmental version, we generally try to lean on simple domain types rather than primitives. so instead of using an int, you'd use a Duration and get the millisecond delay from that. This is obviously more of a challenge (to deserialise from environmental vars etc), but we can also use the policy idea from #3 to develop this in isolation from the behaviours so that can be done later :)

 That was a bit of a braindump as well - if it doesn't make sense then just yell. :)

IgorPerikov:
> if it doesn't make sense then just yell. :)

the points you mentioned are really valuable, I agree with all of them, I'll get back with a new stuff later

IgorPerikov:
A few things I'd like to emphasize here:
- I'm not feeling good about ChaosBehavioursTest, it's really clumsy, especially the one that checks delay, but I decided to show what is the best thing I came up with
- But what is more clumsy is potential test for chaos policies, especially randomized one, I decided not to write test here at all, since it involves a lot of stuff like making a lot of iterations, calculating result distribution, picking right epsilon since nothing is ideal, etc.

IgorPerikov:
@daviddenton kindly ping?

release-maven-central.sh
```diff
@@ -65,5 +65,6 @@ maven_publish "http4k-server-jetty"
 maven_publish "http4k-server-netty"
 maven_publish "http4k-server-undertow"
 maven_publish "http4k-serverless-lambda"
+maven_publish "http4k-testing-chaos"
```
daviddenton: For the moment we need to remove this line from the commit - because there is a cycle for when we create a new module - it needs to be uploaded to bintray, then we have to ask permission for it to be linked to JCenter, and only then can we sync it to maven. If we add this line now, it'll fail build that we release it in! :)

http4k-testing-chaos/src/main/kotlin/org/http4k/chaos/ChaosFilters.kt
```diff
@@ -0,0 +1,34 @@
...
+
+const val LATENCY_HEADER = "X-Chaos-Behaviour"
+
+object ChaosFilters {
+    object ChaosPreFilter {
+        operator fun invoke(chaosPolicy: ChaosPolicy, behaviour: ChaosBehaviour): Filter = Filter { next ->
+            { request ->
+                if (chaosPolicy.shouldInject(request)) {
+                    behaviour.inject(request)
+                    next(request).header(LATENCY_HEADER, behaviour.javaClass.name)
```
daviddenton: This is no longer LATENCY only - just call it X-http4k-chaos or similar. We can also add the description of the behaviour as a method on the behaviour instance, and then use it to generate dynamic names based on what we did! :)

http4k-testing-chaos/src/main/kotlin/org/http4k/chaos/ChaosFilters.kt
```diff
@@ -0,0 +1,34 @@
+package org.http4k.chaos
+
+import org.http4k.core.Filter
+
+const val LATENCY_HEADER = "X-Chaos-Behaviour"
+
+object ChaosFilters {
+    object ChaosPreFilter {
```
daviddenton: Can probably combine these into a single filter and then have 2 methods on the Policy shouldInject(Request) and shouldInject(Response) - both of which can default to false.

daviddenton:
@IgorPerikov sorry for going quiet - given you comments I wasn't sure if you were still actively committing to the branch so left you to it! :)

Anyway - it's a good start - I've left a couple of comments - the other thing that might help to drive out an API (specifically around the configuration) is to introduce further behaviours now:

1. Kill the process!
2. Block the current thread entirely (thread starvation)
3. Send back an HTTP 500.

IgorPerikov:
@daviddenton new stuff is ready to review! 
 
When I was working under it, I realised, that it would be extremely conventional to have two methods for ChaosBehaviour as well - one for request and one for response

latest changelist:
* merge pre and post filters into single one
* add inject(Response) & inject(Request) methods to behaviour interface
* provide more chaos behaviours

delete_version.sh
```diff
@@ -48,5 +48,6 @@ maven_publish "http4k-server-jetty"
 maven_publish "http4k-server-netty"
 maven_publish "http4k-server-undertow"
 maven_publish "http4k-serverless-lambda"
+maven_publish "http4k-testing-chaos"
```
daviddenton: this needs to be removed as we won't publish the module until it's in jcenter..

http4k-testing-chaos/src/main/kotlin/org/http4k/chaos/ChaosBehaviours.kt
```diff
@@ -0,0 +1,69 @@
...
+        fun fromEnv(): AdditionalLatencyBehaviour {
+            val minDelay = Duration.parse(System.getenv("CHAOS_LATENCY_MS_MIN") ?: "PT0.1S")
+            val maxDelay = Duration.parse(System.getenv("CHAOS_LATENCY_MS_MAX") ?: "PT0.5S")
+            return AdditionalLatencyBehaviour(minDelay, maxDelay)
+        }
+    }
+}
+
+class ChaosException(message: String) : Exception(message)
+
```
daviddenton: generally, the pattern within the project is for this kind of thing to appear as an extension function of the parent type. So, for instance `fun ChaosBehaviour.ThrowException()`. This one in particular could take an exception as an arg, and default to throw ChaosException as the default.

http4k-testing-chaos/src/main/kotlin/org/http4k/chaos/ChaosPolicies.kt
```diff
@@ -0,0 +1,23 @@
+package org.http4k.chaos
+
+import org.http4k.core.Request
+import org.http4k.core.Response
+import java.util.concurrent.ThreadLocalRandom
+
+interface ChaosPolicy {
```
daviddenton: ditto the remark about extension functions (from above)

IgorPerikov:
> let me know what you think

sure I will follow the project rules, but tbh I'm not really following this design principle - not a kotlin native speaker yet :). Is this what are you talking here? 
```kotlin
class ExceptionThrowingBehaviour : ChaosBehaviour {
    override val description = "Exception"
    override fun inject(response: Response) = throwException()
}

fun ChaosBehaviour.throwException(exception: Exception = ChaosException("Chaos behaviour injected!")): Response {
    throw exception
}
```
I guess my assumption is wrong, since it seems really weird for me, but I decided to show it just in case :)

At first I thought about moving all the subtypes to be a companion object functions(pretty much like `ContentNegotiation` interface in `body.kt`), and it seems more convenient for me, but AFAIK it has nothing in common with extension functions :open_mouth: 

```kotlin
interface ChaosBehaviour {
    ... // signatures omitted for brevity
    companion object {
        fun throwException(exception: Exception = ChaosException("Chaos behaviour injected!")) = object : ChaosBehaviour {
            override val description = "Exception"

            override fun inject(response: Response): Response {
                throw exception
            }
        }
    }
}
```

In addition to explanation, I'd be really grateful, if you could point me a few places in codebase where I could find an examples of this pattern for better understanding

@daviddenton

daviddenton:
Hi @IgorPerikov - first off, I'm **really, really** sorry for the slow turnaround in this - I've been working up against a deadline so am completely flooded at the moment. It should be better by the end of next week so bear with me! Startups are hard! :)

And whilst I'm waiting for a build to run....

Your intuition was completely right about ContentNegotiation being the template for what I meant - the companion object of the ChaosBehaviour interface would be where we would put these core behaviours. Then further behaviours can be added as extension methods TO the companion object, and hence we keep it all nice and discoverable in the IDE.

This also means that you can drop the name "Behaviour" from each of the implementing classes, so you'd have: ChaosBehaviour.ThrowException**. I'd also maybe give the API use the option to use their own exception and keep the ChaosException as a default parameter in the function

** note the capitalization here - we follow a pattern where you can't really tell if a function is a constructor OR a class. This means that if we move from a function to a class at a later date, it doesn't break the API for the user.

IgorPerikov:
Hi, no need for apologize at all, no matter what is the reason. Thanks for clarification - clear and neat, will come back with update when it's ready

IgorPerikov:
seems like I am already back :) don't hesitate to mark/correct names in the code - I got used to them plus english is not my mother tongue

daviddenton:
Sorry (again) - I totally missed your message that you were done. Good news is that stupid deadline is done, so I can start to concentrate on other things! :)

IgorPerikov:
Glad that you meet your deadline! :fireworks:

PR #138, opened 2018-06-27
Add support for propagation of the Zipkin x-b3-sampled header

jshiell:
As per the [Zipkin spec](https://github.com/openzipkin/b3-propagation) there's a standardised sampling decision header that can be used to inform clients as to whether they should sample the current request. This adds support for the capture and propagation of this header, allowing recording filters to decide whether or not to sample based on an upstream decision.

Note that we default to "sample" so as to maintain backwards compatibility.

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/filter/ZipkinTraces.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/filter/ZipkinTraces.kt
@@ -2,6 +2,8 @@ package org.http4k.filter
 
 import org.http4k.core.HttpMessage
 import org.http4k.core.with
+import org.http4k.filter.SamplingDecision.Companion.SAMPLE
+import org.http4k.filter.SamplingDecision.Companion.from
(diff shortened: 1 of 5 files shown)


jshiell:
On the CI failure:
```
./release-site.sh: line 24: GH_TOKEN: unbound variable
The command "./release-site.sh" exited with 1.
```
Pretty sure that's not me :-D

daviddenton:
Yeah - that's not you. We've tried (and failed) to get release-site.sh NOT to run on a PR build.

You can probably see in it's history those attempts - if you've got any ideas then that'd be ace! :)

Do you need this released sharpish?

jshiell:
I’ll have a look on the script tomorrow. No rush on the release - I’d like to do something with it tomorrow, but SN can pay to have Ivan do the deed 😀 Cheers sir.

s4nchez:
@jshiell It'll be available in 3.32.0. 

(building as we speak, so it'll be available soon in jCenter, and Maven central after that)

PR #140, opened 2018-06-28
Changed doc invocation via YML to avoid PR builds

jshiell:
Not sure if `TRAVIS_PULL_REQUEST` is exported to the subshell - in any case, this is how Travis themselves recommend accomplishing this.

Here goes nothing...


daviddenton:
That works. We still need to make the same change to the API building, but I can do that. Thanks!

Issue #143, opened 2018-07-09
Request.form(name) reparses body every invocation?


dmcg:
I've just been reviewing some of my code

```kotlin
Request.toUnsavedCustomer(): Result<UnsavedCegidCustomer> = resultOf {
    UnsavedCegidCustomer(
        salesPersonId = requiredForm("associateId"), // See *
        id = null,
        titleId = form("title").validatedAsTitle(),
        firstName = requiredForm("firstName"),
        lastName = requiredForm("lastName"),
        dateOfBirth = form("dateOfBirth").nullIfBlank().toLocalDate(),
        email = form("email").nullIfBlank(),
        phone = form("phone").nullIfBlank(),
        mobilePhone = form("mobilePhone").nullIfBlank(),
        wechatId = form("wechatId").nullIfBlank(),
        nationalityIso3 = form("nationalityISO3").nullIfBlank(),
        addressLine1 = form("addressLine1").nullIfBlank(),
        addressLine2 = form("addressLine2").nullIfBlank(),
        addressLine3 = form("addressLine3").nullIfBlank(),
        city = form("city").nullIfBlank(),
        regionId = form("regionId").nullIfBlank(),
        countryIso3 = form("countryIso3").nullIfBlank(),
        zipCode = form("zipCode").nullIfBlank(),
        postOptIn = form("postConsent").isChecked(),
        emailOptIn = form("emailConsent").isChecked(),
        phoneOptIn = form("phoneConsent").isChecked(),
        appsOptIn = form("appsConsent").isChecked()
    )

private fun Request.requiredForm(name: String) = form(name).let {
    when {
        it != null && it.isNotBlank() -> it
        else -> error("Missing $name")
    }
}
```

and realise that every one of those Request.form(name) invocations is reparsing the body?

I don't expect HTTP4K to cache the parsed form, but it does seem that the API is encouraging me to do a nasty thing

daviddenton:
Well TBH it's a balance between convenience and simplicity. If you want to avoid the lenses style, in this case you can call request.form() to get a list of parsed fields and then go from there.

We could add a note to the doc (PR it?) - don't really want to deprecate.

Thoughts ?

dmcg:
It's made more difficult by Parameters.findSingle being internal, and as I look at it, every find is a find over a list rather than a map. Does the lenses do a more efficient job?

dmcg:
Thanks Ivan - not a bad compromise I think

daviddenton:
[...]

PR #144, opened 2018-07-10
Conversions between Parameters and map

dmcg:
https://github.com/http4k/http4k/issues/143


s4nchez:
@dmcg Thank you!

That's going through the pipeline and should be available in version 3.33.0

PR #166, opened 2018-08-24
JSON-RPC module

kirderf:
This PR adds a new module to support JSON-RPC services over HTTP: `http4k-jsonrpc`

## Features

- Fully compliant with version 2.0 of the JSON-RPC specification: http://www.simple-is-better.org/json-rpc/jsonrpc20.html
  - Version 1 JSON-RPC requests are not supported.
  - Supports:
    - "normal" and notification request and response
    - batch request and response
    - error objects with custom error code, message and optional error data
- Implements tunnelling of JSON-RPC requests over HTTP, as mentioned in this proposal: http://www.simple-is-better.org/json-rpc/transport_http.html
  - `POST` requests are implemented according to proposal
  - `GET` requests are not allowed, as recommended
  - HTTP response status codes map to proposal
- Enforces the user to use command objects (read kotlin data classes) for method invocations. This makes it easier to do things like parameter validation, parameter field mappings, implement a CQRS architecture, etc.
- Supports both manual and automatic request/response payload mapping. Auto mapping is only allowed if the underlying JSON module used extends `JsonLibAutoMarshallingJson`, e.g. Jackson. Manual mapping is done using lenses, modelled as `Param` and `Result` respectively.
- Both mapping styles allows the user to map array parameter payloads to the correct fields.
- Custom error handling, where the user can map specific exceptions to JSON-RPC error objects.
- Only depends on `http4k-core` and one of the `http4k-format-xxx` modules.

## Examples

Given the following simple service:
```kotlin
data class Add(val first: Int, val second: Int)
data class Div(val dividend: Double, val divisor: Double)

class DivideByZeroException(val dividend: Double) : RuntimeException("divide by zero")

object Calculator {
    fun add(it: Add): Int = it.first + it.second

    fun divide(it: Div): Double = if (it.divisor == 0.0 ) {
        throw DivideByZeroException(it.dividend)
    } else {
        it.dividend / it.divisor
    }
}
```

**To set up a JSON-RPC service, using manual mapping:**
```kotlin
val handler: HttpHandler = "/calculator" bind jsonRpc(Jackson) {  // or jsonRpc(manual(Jackson)) {
    
    val addParams = Params<JsonNode, Add> {
        Add(it["first"].asText().toInt(), it["second"].asText().toInt())
    }
    val addResult: Result<Int, JsonNode> = Result { json.number(it) }

    val divParams = Params<JsonNode, Div> {
        Div(it["dividend"].asDouble(), it["divisor"].asDouble())
    }
    val divResult: Result<Double, JsonNode> = Result { json.number(it) }
    
    method("add", handler(setOf("first", "second"), addParams, addResult, Calculator::add))
    method("divide", handler(setOf("dividend", "divisor"), divParams, divResult, Calculator::divide))
}
```
Any exceptions thrown in the `Param` lenses, are caught and mapped to the correct error object (code `-32602 : Invalid params`). Any exceptions thrown when invoking the service method are caught and mapped to error object `-32603 : Internal error`.

**To use auto mapping, the same service mapping would look like this:**
```kotlin
val handler: HttpHandler = "/calculator" bind jsonRpc(auto(Jackson)) {
    method("add", handler(setOf("first", "second"), Calculator::add))
    method("divide", handler(setOf("dividend", "divisor"), Calculator::divide))
}
```

**If the user does not want to support array style parameters, then the service mapping would look like:**
```kotlin
val handler: HttpHandler = "/calculator" bind jsonRpc(auto(Jackson)) {
    method("add", handler(Calculator::add))
    method("divide", handler(Calculator::divide))
}
```

**A custom error handler can be defined to handle `DivideByZeroException` or any other exceptions like this:**
```kotlin
private object CalculatorErrorHandler : ErrorHandler {
    override fun invoke(error: Throwable): ErrorMessage? = when (error) {
        is DivideByZeroException -> DivideByZeroErrorMessage(error.dividend)
        else -> null
    }

    private class DivideByZeroErrorMessage(private val dividend: Double) : ErrorMessage(1, "Divide by zero") {
        override fun <ROOT : NODE, NODE> data(json: Json<ROOT, NODE>): NODE? =
                json.string("cannot divide $dividend by zero")
    }
}

val handler: HttpHandler = "/calculator" bind jsonRpc(auto(Jackson), CalculatorErrorHandler) {
    method("add", handler(Calculator::add))
    method("divide", handler(Calculator::divide))
}
``` 
This example also shows how to include a custom `data` field in the error message object, which can be of any valid json structure or primitive.

## Not included
- A JSON-RPC version 2 compliant client abstraction; this PR only focusses on services.
- No user documentation has been included. The info in this PR can be used as the basis for such documentation.


kirderf:
Looking at the report from Codebeat, it's clear that some improvements to the code can be done. Want me to fix those?

daviddenton:
Thanks for the PR :) 

I'm totally happy to merge it now and then we can have a bit of a further play to see if there are improvements to be made. TBH, I have real difficultly reading code from Diffs, so find it easier to play with that stuff in the IDE and make suggestions from there.

http4k-jsonrpc/src/main/kotlin/org/http4k/jsonrpc/ErrorMessage.kt
```diff
@@ -0,0 +1,29 @@
...
+    companion object {
+        val InvalidRequest = ErrorMessage(-32600, "Invalid Request")
+        val MethodNotFound = ErrorMessage(-32601, "Method not found")
+        val InvalidParams = ErrorMessage(-32602, "Invalid params")
+        val InternalError = ErrorMessage(-32603, "Internal error")
+        val ParseError = ErrorMessage(-32700, "Parse error")
+    }
+}
+
+interface JsonNodeProducer {
```
daviddenton: Is this interface needed? looks like there is only one usage

kirderf: No I guess not. Could simply be incorporated in to the ErrorMessage class. I was going to use that interface in other domain classes, but didn't really plan out that way.

daviddenton: I've removed it anyway in tweaks I've made to the code already - so might be worth you rebasing from master to get those. :)

daviddenton: Couple of other things. There are a few unused methods in extensionsKt. Are these required - or at least maybe meant to be covered by a test?

Also, for some reason coveralls is complaining that there is 0% coverage on the new module, which is a bit annoying as it's dragging down the percentage score. :(  If you could have a look to see what might be missing that'd be ace.

daviddenton: Np
 - I'm away now as well! 

Take a look at the other branch. And let me know what you think. I made some tweaks to hide the construction of the builder.

kirderf: Ah ok. I'll have a more detailed look and let you know what I think. Initial feedback is that it look ok to me. The reason I chose to use extension methods was that I wanted to base it on what was done for the Contract module. Using methods on an object might make more sense though and makes it more explicit.

kirderf: LGTM. Created another PR for some minor changes to that branch: https://github.com/http4k/http4k/pull/174

PR #169, opened 2018-08-29
Make some useful things public

dmcg:
Blimey, I contribute the stuff and then you take it away from me!

InMemoryResource is a convenience by anyone wanting to implement their own resource serving.

ResourceListingHandler is needed if you want to implement your own listings (I do!)


daviddenton:
ok - ok - stop whinging!

PR #170, opened 2018-08-29
allow navigation to Uri objects

dickon:
In master if you do:

`driver.navigate().to(Uri.of("http://slashdot.org")) `

then you'll find that to is resolved to an extension method to construct pairs in the Kotlin standard library which will produce code with no effect, which could be confusing.


daviddenton:
Thanks! :)

PR #171, opened 2018-08-29
allow empty action items

dickon:
I found a form in the wild which had an action attribute with no value. This change tests and handles that case in the WebDriver. I've followed the pattern of having callback parameters from the Element that can be used with the driver, rather than keep a WebDriver reference in all SoupElements. What's unfortunately touches more files than I'd like.

I can extend the pattern of counting the number of times the HTTPHandler gets called into the other test cases if you like.


dickon:
While I think this is good to merge as it is I'm currently building on top of the change to handle getting values from buttons and submitting based on sending click messages to buttons..

dickon:
Okay, now added logic and tests to handle buttons. While I'd like to find a better way to code some the submit() method I think this is correct and useful so please merge.

PR #176, opened 2018-09-03
improve web driver support for forms

dickon:
- detect buttons if they are clicked on
- avoid losing host name/protocol/authority for form submissions
- check a couple of different way actions can be specified
- add negative equality test on elements since we rely on that in the form code


s4nchez:
Thank you for the improvement, @dickon!

Issue #177, opened 2018-09-03
In-memory request context store should be thread-safe?


zvozin:
https://github.com/http4k/http4k/blob/3170cc58d0bd50e777cb21a101639e678ca300a1/http4k-core/src/main/kotlin/org/http4k/core/RequestContexts.kt#L11 - that's a `LinkedHashMap<A, B>` in what appears to be a multi-thread-access situation. Would a `ConcurrentHashMap<A, B>` be more suitable?

Asking cause https://github.com/http4k/http4k/blob/3170cc58d0bd50e777cb21a101639e678ca300a1/http4k-core/src/main/kotlin/org/http4k/core/RequestContext.kt#L9 makes `id` internal, which makes rolling your own impossible.

daviddenton:
Sounds reasonable - if you want to fix and PR those 2 changes then that's be great! Definitely want people to be able to implement their own store.

zvozin:
https://github.com/http4k/http4k/pull/178

daviddenton:
Fixed by #178

PR #178, opened 2018-09-04
https://github.com/http4k/http4k/issues/177

zvozin:


zvozin:
Unless I'm misreading the build, log the test failure doesn't appear to be caused by the PR.

zvozin:
@daviddenton what's the preferred way to know when a new release rolls out?

PR #179, opened 2018-09-13
Always execute jsonrpc method even if notification request

kirderf:


kirderf:
I'll fix the Coveralls issues.

PR #187, opened 2018-10-16
 gradle-kotlin-dsl-libs version "0.2.4"

jmfayard:
Changelog:

- #1 Make it slightly easier to manually update the version
- #3 Add buildSrc/.gitignore
- #4 SyncLibs generate wrong output file


s4nchez:
This change is fine. 

The only thing strange is the change to MimeUtility. That's a copy from apache commons, so I think it should be touched if possible. Or am I missing something?

jmfayard:
@s4nchez you are totally right, I didn't intended to do this change. 
Was probably an automatic cleanup from intellij. It's reverted now

PR #192, opened 2018-10-25
Use UTC when checking cookie expiry

dmgd:
so that parsed expiry dates are not affected by system timezone (since Clock.instant() is always in UTC)


s4nchez:
@dmgd Thank you for spotting and fixing it!

That's now merged and released in 3.94.1

PR #195, opened 2018-11-02
Add filter allowing Gzipping based on an allowed set of content types

jshiell:
The current GZip filter is an all-or-nothing affair. This extends it so that you can give it a set of content-types which will restrict the responses for which GZipping will be performed. This allows you to ensure that previously compressed content such as PNGs will not be recompressed.

I also fixed a Javadoc, as you lot are hacks who do too much copy & paste ;-)


s4nchez:
Thanks for the patch, @jshiell ! I’ll try and get it released ASAP.

s4nchez:
That'll be out in 3.98.0 (being released as we speak)

jshiell:
Cheers @s4nchez  - as always, you're a saint 👼

PR #204, opened 2019-01-03
Include session token in AwsCredentials

dhobbs:
Added optional sessionToken to AwsCredentials and include it in the signed request if present


daviddenton:
Intending to take this, but then to add the ability to pass in a creds provider function to the aws filter, so people can implement dynamic creds without having to rebuild the filter stack

dhobbs:
I'm happy to do this (because I need it!). Do you want the original signature left intact, or can I replace the `credentials: AwsCredentials` argument with `() -> AwsCredentials`? This would obviously break a few things...

PR #205, opened 2019-01-04
Preventing FallbackCacheControl from duplicating existing headers

leandronunes85:
FallbackCacheControl claims it will include a default value for "Cache-Control", "Expires" and "Vary" headers when they are not present in the response. However it is actually duplicating the existing values when these headers are already present in the response. This PR fixes that.


daviddenton:
Thanks! :)

PR #207, opened 2019-01-14
Issue with prefixes in static routing

ruXlab:
Hello here

I found a case when paths are not translated properly

For example request `GET /bar/bar-xyz.html` to `routes("/bar" bind static(Classpath("bar")))` was translated into`/-xyz.html`


daviddenton:
Thanks for the fix @ruXlab  😄

PR #227, opened 2019-04-24
Ensure dir for linkmapping in dokka task is a unix path.

JordanBergin:
When importing the latest http4k in Intellij on Windows it fails with message:

> Incorrect dir property, only Unix based path allowed.

This is happening in the dokka task because the dir property in the linkmapping is being passed a Windows file path, but since dokka 0.9.18 this isn't allowed (verified by trying 0.9.17 and the project imported fine).

I believe Replacing whatever file separator the OS uses with "/" will always result in a valid Unix path for the projectDir in this case.

Only tested on my Windows machine since it's such a small change but I'm happy to spin up different OS and verify there too if need be.


s4nchez:
Thank you for spotting and fixing this, @JordanBergin :)

PR #252, opened 2019-06-09
Made the multipart header parser case-insensitive.

tenniscp25:
We build our mobile client with react-native. We use plain XMLHttpRequest and FormData to create file upload requests. We found that http4k failed to parse lowercase multipart header field names generated by XMLHttpRequest.


s4nchez:
Hi @tenniscp25 , thank you for spotting this and the PR.

Would you mind providing some background on how you came across the need for this change or, even better, add a test case in `MultipartFormTest` or `StreamingMultipartFormHappyTests` so it gets checked and documented?

tenniscp25:
Gladly. I've updated the PR comment and added the test.

s4nchez:
Thank you, @tenniscp25 

Your change is being released as we speak and will be available on version 3.154.1

tenniscp25:
Thank you for the quick release!

Issue #250, opened 2019-06-08
Documentation: Move examples from dependencies { compile } to dependencies { implementation }?


ivoanjo:
Hello there, and a big thanks for http4k!

I was bootstrapping a new project, and saw that the project documentation still recommends using `dependencies { compile ... }` to import http4k as a dependency, whereas the [gradle documentation](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation) now mentions that

> The `compile` configuration still exists but should not be used as it will not offer the guarantees that the `api` and `implementation` configurations provide.

[This](https://stackoverflow.com/questions/44493378/whats-the-difference-between-implementation-and-compile-in-gradle) stackoverflow link is also relevant.

Would you accept a PR to update the documentation to use `implementation` instead? I don't mind tackling it, but wanted to know if there may be a reason for still wanting to have the old syntax :)

s4nchez:
I’m not aware of any good reason to keep the old syntax. A PR to change would be welcome 🙂

daviddenton:
@ivoanjo thanks very much for this - it's  great to stay on top of the tech, and just something that we haven't got some to.

One thing though - I think we should be consistent across the project, so we should also be looking at converting our project gradle files to the new format as well.

It doesn't have to be this PR or issue, so maybe let's raise another one to cover it?  (although if you wanted to do it as part of your PR then that'd be amazing 😉).

ivoanjo:
@daviddenton sure, I can take a stab at it!

I usually favor smaller PRs, but I can add to that one if you prefer :)

daviddenton:
Whichever you'd like to do would be absolutely brilliant. 🙃

PR #254, opened 2019-06-13
Modernize gradle dependency configuration

ivoanjo:
The current version of gradle recommends not using the `compile`
dependency configuration:

> The `compile` configuration still exists but should not be used as it
> will not offer the guarantees that the `api` and `implementation`
> configurations provide.

(from <https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation>)

Thus, I've changed every `testCompile` to `testImplementation` (minor),
and more importantly, every `compile` to `api`.

Why `api`, rather than `implementation`?

> Dependencies appearing in the `api` configurations will be
> transitively exposed to consumers of the library, and as such will
> appear on the compile classpath of consumers.
> Dependencies found in the `implementation` configuration will, on the
> other hand, not be exposed to consumers, and therefore not leak into
> the consumers' compile classpath.

Because the old `compile` dependency configuration behaved as the `api`
one, in that it exposed the dependencies of http4k and its modules
to consumers, changing it now to `implementation` would be a breaking
change -- as a user updating http4k and accidentally referencing
one of its dependencies in its code would suddenly get errors about
missing dependencies.

Instead, from now on, new dependencies should be evaluated if they
should be added as `api` or `implementation`, and in a future 4.x
version of http4k, the current items marked as `implementation` can
be revisited.

Issue #250

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -226,7 +226,7 @@ dependencies {
     provided project(":http4k-testing-hamkrest")
     provided project(":http4k-testing-webdriver")
 
-    testCompile Config.TestDependencies
+    testImplementation Config.TestDependencies
 }
 
 sourceSets {
--- a/http4k-aws/build.gradle
+++ b/http4k-aws/build.gradle
@@ -4,7 +4,7 @@ dependencies {
     provided Libs.kotlin_stdlib_jdk8
     provided project(":http4k-core")
 
-    testCompile project(path: ":http4k-core", configuration: 'testArtifacts')
-    testCompile project(":http4k-client-apache")
-    testCompile Config.TestDependencies
+    testImplementation project(path: ":http4k-core", configuration: 'testArtifacts')
+    testImplementation project(":http4k-client-apache")
+    testImplementation Config.TestDependencies
 }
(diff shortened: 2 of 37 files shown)


daviddenton:
Thanks VM for the PRs to fix this stuff up. 😄

daviddenton:
just an FYI - I've had to roll back the change for implementation -> api change. We just released http4k v3.155.1, upgraded it in our current project and everything went red. I'm in the middle of something that needs one of the changes in the new release, so no time to look at it right now.

ivoanjo:
Apologies @daviddenton for the issue!

I'm a bit new around gradle too, let me know if you have any hints on how I can debug this :)

daviddenton:
np - it was a (somewhat admitedly nasty) surprise to us as well! :)

I *think* the problem may be that this new mechanism will only work if the project that you're importing it to is also using the new api/implementation mechanic. Our target project is using the old `compile` mechanic, and basically it didn't actually see any of the transitive dependencies, which is obviously really bad from a user perspective if they have to upgrade their project files just to get a new version of http4k.

I could be wrong though.. :)

PR #255, opened 2019-06-14
Improve api documentation: Link to JDK 9 and dependencies' javadocs

ivoanjo:
I've configured dokka to target JDK 9 (later versions not possible due to bug, see commit for more details), and also added links to many of http4k dependencies' javadocs, so that they are usable from the api docs.


daviddenton:
Thanks! :)

PR #264, opened 2019-07-12
Updated error message for when a request has not been routed

liam056:
Removed the double negative from the error message.


daviddenton:
Thanks for the contribution 😀

PR #266, opened 2019-07-16
Allowing openapi v3 to cope with json naming annotations

tom:
Not sure if this is the final solution, but it should illustrate the problem


tom:
I've added a public composite, I'm wondering if leaving it would be better, as I suspect more annotations might be needed in the future that way it's easy to add them as required, and it won't break things for exciting people

daviddenton:
Cool. Or even existing people, excited though they may be to be using such amazing code. 😉

Get @s4nchez to release it for you if you need it urgently.

PR #267, opened 2019-07-17
Rationalise ContractRoute to satisfy HttpHandler

rgladwell:
`ContractRoute` does not implement `HttpHandler` making it non-obvious how to test contract routes in isolation. This change makes this class more uniform and makes testing easier.


daviddenton:
As discussed previously - I've implemented this in a similar way but a little bit different. Have given you a credit on the CHANGELOG.

Thanks for the inspiration! 😄

PR #268, opened 2019-07-22
Add support for the OpenAPI2 host directive

rgladwell:
Added support for the OpenAPI2 host directive:

https://swagger.io/docs/specification/2-0/api-host-and-base-path/

This property is required by Google Cloud Endpoints:

https://cloud.google.com/endpoints/docs/openapi/configure-endpoints

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRenderer.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRenderer.kt
@@ -4,6 +4,5 @@ import org.http4k.contract.security.Security
 import org.http4k.core.Response
 
 interface ContractRenderer : ErrorResponseRenderer {
-    fun description(contractRoot: PathSegments, security: Security, routes: List<ContractRoute>, host: String): Response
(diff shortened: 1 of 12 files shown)


daviddenton:
Happy to add this, but a couple of differences on the implementation I think we would want:

1. Since the "host" value is specific to the OpenApi2, this value would be better passed as a value on the OpenApi2 renderer object and not in the general contract description block. If we add "servers" support to OA3, this would also be added on the renderer as well.
2. Would like to stay away from `String` if we can - since this has a port as well I think we'd prefer to model this using a `Uri?` (defaulting to null), and then use `uri.authority` to get the value. Happy to discuss this one a little further.. 😄

rgladwell:
@daviddenton fixed in commit eb76bfebcec3cf517725151f550994451b500ee5. Let me know what you think.

daviddenton:
yep - looks better. There is one neat trick which we can do to remove the propertiesWithUriData null check, but I'll do that after merge. Thanks! 😄

rgladwell:
👍

PR #269, opened 2019-07-22
Fix invalid OpenAPI2 when root and base path match

rgladwell:
When generating OpenAPI2 descriptions where the root and base path are the same, an invalid description is generated.

For example:

```kotlin
  val router = "/" bind contract {
      routes += "/" bindContract Method.GET to { Response(Status.OK) }
      descriptionPath = "/docs"
  }
```

Would generate swagger as follows:

```json
  {
    "swagger": "2.0",
    "info": {
      "title": "title",
      "version": "1.2",
      "description": "module description"
    },
    "basePath": "/",
    "paths": {
      "": {
        "get": {
          "operationId": "get"
        }
      }
    }
  }
```

Which is invalid as empty strings for `#path` URIs are not allowed. This can be tested by running the output of the API through a swagger validator as follows:

```sh
  $ npm install swagger-cli -g
  $ swagger-cli validate example-http4k-swagger.json
```


daviddenton:
thanks for the contrib :)

daviddenton:
One teenytiny thing to note for future - we generally try to never use * imports in the code. It's not a massive problem as we can just pick it up after the merge, but it makes reviewing easier 😄

PR #270, opened 2019-07-24
Optionally include description path in description

rgladwell:
Google Cloud Endpoints (GCE) will only make an endpoint publicly accessible if it has a path directive in the OpenAPI description. But by default http4k-contract does not include the contract path in the OpenAPI description.

This means hosted Swagger documentation is inaccessible unless we manually re-format the results of the contract route.

This change adds an optional directive to the contract DSL that adds the description path to the OpenAPI description, allowing it to be accessed when deployed in GCE.


rgladwell:
See this thread for more information:

https://kotlinlang.slack.com/archives/C5AL3AKUY/p1563460340164100

PR #274, opened 2019-08-12
Add support for Google Cloud Endpoints OAuth2

rgladwell:
This ticket adds support for Google Cloud Endpoints (GCE) OAuth2 as defined in this document:

https://cloud.google.com/endpoints/docs/openapi/authenticating-users-custom

Note: by default the GCE Extensible Service Proxy handles authentication, so the new GCE Security implementation sets a NOOP authentication filter by default.

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v2/OpenApi2SecurityRenderer.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v2/OpenApi2SecurityRenderer.kt
@@ -6,11 +6,12 @@ import org.http4k.contract.openapi.SecurityRenderer
 import org.http4k.contract.openapi.rendererFor
 import org.http4k.contract.security.ApiKeySecurity
 import org.http4k.contract.security.BasicAuthSecurity
+import org.http4k.contract.security.ImplicitOAuthSecurity
 
 /**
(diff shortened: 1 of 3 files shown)


daviddenton:
I think for this we should change http4k to support the implicit flow first, and then from there we can provide a custom "preconfigured instance" for GCE (a la the OAuthProviders). We'll have a think about how to make these changes so that there is the minimum disruption, as currently the normal OAuthSecurity is hardwired to the AuthCode flow.

rgladwell:
@daviddenton thanks for your feedback, updated the PR with your comments. Please let us know what you think.

rgladwell:
@daviddenton Looks like the build is failing because the `flow` field in an OAuth2 security descriptor is required in OpenAPI2. Not sure how to proceed.

daviddenton:
Ok - I've remodelled the Security stuff to include the implicit flow and a googleCloudEndpoints definition - but I've done it for V3 only. Could you update this to reapply the support to V2 - not sure if the format is different.

rgladwell:
@daviddenton cool, thanks, we'll rebase and integrate with your changes.

rgladwell:
@daviddenton done, let me know if it is OK, please!

daviddenton:
releasing this now as 3.173.0

PR #277, opened 2019-08-19
Multiple global security definitions

rgladwell:
Add support for multiple global security definitions, reducing duplication and exposure to security issues.

Currently `http4k-contract` if you need more than one global security definition you have to specify one global and multiple single `Security` implementations for each endpoint. This can be verbose leading to ugly Swagger documentation, such as follows:

```js
	"securityDefinitions": {
		"oauth2-googleid": {
			"type": "oauth2",
			"flows": "implicit",
			"authorizationUrl": "",
			"x-google-issuer": "https://accounts.google.com",
			"x-google-audiences": "client@example.org"
		},
		"oauth2-service": {
			"type": "oauth2",
			"flows": "implicit",
			"authorizationUrl": "",
			"x-google-issuer": "service-client-id@example.org",
			"x-google-jwks_uri": "https://www.googleapis.com/robot/v1/metadata/x509/service-client-id@example.org"
		},
		"oauth2-service-2": {
			"type": "oauth2",
			"flows": "implicit",
			"authorizationUrl": "",
			"x-google-issuer": "service-client-id@example.org",
			"x-google-jwks_uri": "https://www.googleapis.com/robot/v1/metadata/x509/service-client-id@example.org"
		},
...
```

It is also error prone, as developers may forget to define the additional security definitions for each endpoint.

This change allows additional global security definitions to be defined without breaking backwards compatibility. Unfortunately, this requires a new `securities` directive in the contract DSL which duplicates (in some ways) the original `security` directive.


rgladwell:
Closed in favour of [multiple-securities](https://github.com/http4k/http4k/commits/multiple-securities) branch.

PR #288, opened 2019-09-05
Make client_secret optional AuthorizationCodeAccessTokenRequest to be…

tom:
…tter support non client_secret flows


http4k-security-oauth/src/main/kotlin/org/http4k/security/oauth/server/accesstoken/AuthorizationCodeAccessTokenGenerator.kt
```diff
@@ -78,7 +78,7 @@ private object AuthorizationCodeAccessTokenForm {
         with(accessTokenForm(request)) {
             AuthorizationCodeAccessTokenRequest(
                 clientId(this),
-                clientSecret(this),
+                clientSecret(this) ?: "",
```
daviddenton: doesn't this want to be a nullable now?

tom: Maybe, I followed the pattern from places like this

ClientSecretAccessTokenRequestAuthentication

Though I think .orEmpty will be better so will do that

PR #290, opened 2019-09-09
Add support for AWS Lambda context + request

### LINKED ISSUE #287: Expose the AWS Lambda context to the handlers
Hello, I am trying to use the [Http4k serverless](https://www.http4k.org/guide/modules/serverless/) feature for some of my Lambda functions, but I need to get access to the [`AWS Lambda Java Context object`](https://docs.aws.amazon.com/lambda/latest/dg/java-context-object.html).

Based on the [Java Lambda function handler docs](https://docs.aws.amazon.com/lambda/latest/dg/java-programming-model-req-resp.html) just having a second argument to the Lambda function entrypoint is enough to give you the context.

However, the [Lambda function `handle`](https://github.com/http4k/http4k/blob/master/http4k-serverless-lambda/src/main/kotlin/org/http4k/serverless/lambda/LambdaFunction.kt#L19) method does not receive the context so essentially it gets lost and I cannot access it from inside my handlers.

I thought of several approaches on getting access to the context so I would like to see if it's possible to get something that can be merged into the project so that others can use it as well, or at least you can tell me if I am missing something that would make it easier for me.

### Approach 1

Create a new `LambdaFunctionWithContext` class which is the entry point to my lambdas and it essentially looks something like:

```
import org.http4k.serverless.lambda.LambdaFunction

class LambdaFunctionLambdaFunctionWithContext(env: Map<String, String> = System.getenv()) {
    fun handle(request: ApiGatewayProxyRequest, context: com.amazonaws.services.lambda.runtime.Context) = {
        GlobalSingleton.set(context)
        LambdaFunction(env).handle(request)
    }
}
```

And from inside my handlers I can do `GlobalSingleton.get()` and retrieve the function context. However, I don't like having global state in my code so I would treat this as a last-way approach, but it is the easiest (and has no concurrency issues due to Lambda's execution model)!

### Approach 2

Fork the [`LambdaFunction`](https://github.com/http4k/http4k/blob/master/http4k-serverless-lambda/src/main/kotlin/org/http4k/serverless/lambda/LambdaFunction.kt) class into a `LambdaFunctionWithContext` and instead of creating a [`Request`](https://github.com/http4k/http4k/blob/5a3dcde40dec1e71ec583c8738532b2bdc9eb574/http4k-core/src/main/kotlin/org/http4k/core/http.kt#L209-L213) object it will create a class `RequestWithContext` that implements `Request` but also contains the AWS Lambda Context and allows the handlers to cast the received `Request` into `RequestWithContext` and extract the context they need.

The downside of this is that basically I stop using your `LambdaFunction` and any improvement you make on it won't be applicable to my lambda functions since I am using a fork copy of it.

### Approach 3

At first when I saw the [`Request Contexts`](https://www.http4k.org/cookbook/request_context/) I thought that it's something that each request handler could use to receive implicit context but it seems that it's not the case and the contexts need to be explicitly given to the handlers.

Is there any way that I missed where the `RequestContexts` can be used and make what I want to achieve above easier than the two approaches described?

If there is an approach that can be merged back into the package I am happy to contribute it, but I think deciding the best approach is the first step towards that.

Thanks a lot for any tips or hints.

ivoanjo:
Following the discussion in #287 , I've gone ahead and used the `RequestContexts` mechanism to expose the underlying AWS Lambda context object, as well as the request object, so that lambdas using http4k have full access to the metadata and tools that AWS Lambda sends along on every invocation.

I've also did a few tweaks to fix Java 8 support, as I'm still stuck on it :(

One important detail is that for backwards-compatibility I did not change the existing `AppLoader` interface, but added `AppLoaderWithContexts`. This way, existing users of the serverless module will not suddenly see compilation errors upon updating.

See the individual commit messages for more details.

I went a bit beyond what was discussed in #287 and also pulled in the `aws-lambda-java-events` library -- my reasoning is documented in the commit message of 
5abbd75:

> The format of the input message for a lambda is documented in
<https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format>
and several of these fields were missing from the previous
`ApiGatewayProxyRequest`/`ApiGatewayProxyResponse`.
>
> This means that in particular, fields such as the `requestContext`,
and `stageVariables`, as well as multi-value headers and parameters
were unavailable.
>
> Rather than manually adding each of the missing fields (and its
nested fields) to the proxy request/response classes, instead I
went ahead and added a dependency on the small (around 100kb as
of this writing --
<https://mvnrepository.com/artifact/com.amazonaws/aws-lambda-java-events/2.2.7>)
`aws-lambda-java-events` library, which already includes POJOs for
all of the fields and their nested objects.

Fixes #287


ivoanjo:
P.s.: I haven't yet tested all of the changes in a real lambda (just testing locally), but otherwise this is feature-complete.

http4k-serverless-lambda/src/main/kotlin/org/http4k/serverless/lambda/LambdaFunction.kt
```diff
@@ -1,31 +1,66 @@
...
+import org.http4k.core.RequestContexts
 import org.http4k.core.Response
 import org.http4k.core.Uri
+import org.http4k.core.then
 import org.http4k.core.toUrlFormEncoded
+import org.http4k.filter.ServerFilters
 import org.http4k.serverless.BootstrapAppLoader
 
+const val LAMBDA_CONTEXT_KEY = "HTTP4K_LAMBDA_CONTEXT"
+const val LAMBDA_REQUEST_KEY = "HTTP4K_LAMBDA_REQUEST"
```
lambrospetrou: Maybe rename this to `LAMBDA_REQUEST_EVENT_KEY`?

http4k-serverless-lambda/src/main/kotlin/org/http4k/serverless/lambda/LambdaFunction.kt
```diff
@@ -1,31 +1,66 @@
...
  * to instantiate the Http4k handler which can be used for further invocations.
  */
 class LambdaFunction(env: Map<String, String> = System.getenv()) {
-    private val app = BootstrapAppLoader(env)
+    private val contexts = RequestContexts()
+    private val app = BootstrapAppLoader(env, contexts)
+    private val initializeRequestContext = ServerFilters.InitialiseRequestContext(contexts)
 
-    fun handle(request: ApiGatewayProxyRequest) = app(request.asHttp4k()).asApiGateway()
+    fun handle(request: APIGatewayProxyRequestEvent, lambdaContext: Context? = null) =
```
lambrospetrou: Is the `Context? = null` here just for the tests? AWS Lambda will always provide this when the function has the second argument.

ivoanjo: Yes. Some of the tests directly use invoke this without setting up the contexts, but on lambda this will indeed always be there.

ivoanjo:
> P.s.: I haven't yet tested all of the changes in a real lambda (just testing locally), but otherwise this is feature-complete.

I've tested it now on a real lambda and it's working correctly:

```kotlin
        val extractLambdaInformation: (Request) -> String = { request: Request ->
            if (contexts != null) {
                val lambdaContext: Context? = contexts[request][LAMBDA_CONTEXT_KEY]
                val lambdaRequest: APIGatewayProxyRequestEvent? = contexts[request][LAMBDA_REQUEST_KEY]
                "\nlambdaInformation: lambdaContext='${lambdaContext?.toString()}'' lambdaRequest='${lambdaRequest?.toString()}'"
            } else {
                ""
            }
        }


        val app: HttpHandler = routes(
            "/ping" bind GET to { request: Request ->
                Response(OK).body("pong! ${extractLambdaInformation(request)}")
            },
```

output:

```
pong! 
lambdaInformation: lambdaContext='lambdainternal.api.LambdaContext@57d5872c'' lambdaRequest='{resource: /ping,path: /ping,httpMethod: GET,requestContext: {accountId: ...,resourceId: tagsi1,stage: test-invoke-stage,requestId: 6c364d35-bb68-4d13-9641-db6e9a184571,identity: {...userAgent: aws-sdk-java/1.11.590 Linux/4.9.184-0.1.ac.235.83.329.metal1.x86_64 OpenJDK_64-Bit_Server_VM/25.212-b03 java/1.8.0_212 vendor/Oracle_Corporation},resourcePath: /ping,httpMethod: GET,apiId: ...,path: /ping,},isBase64Encoded: false}'
```

Let me know what you think!

daviddenton:
Thanks VM for your effort on this. We'll try to get around to looking at it ASAP - it's a bit disconnected from what we're up to at the moment and we'd like to give it due attention, so bear with us. 🙃

http4k-serverless-lambda/src/main/kotlin/org/http4k/serverless/lambda/LambdaFunction.kt
daviddenton: I've merged it but a couple of questions/changes:

1. Instead of putting the entire `APIGatewayProxyRequestEvent` I'm going to put the use the `ProxyRequestContext` to avoid duplication.

daviddenton:
@ivoanjo Just to give you a bit of a heads up that we're intending to remove the dependencies on the amazon JARs in an upcoming `http4k-serverless-lambda` release and to replace them with simple Kotlin data classes instead. There will be no functional impact, but it will break any code that currently uses those classes.

ivoanjo:
Thanks for the heads up @daviddenton! One thing I've since ran into is that in some situations we want to access fields that are not in the amazon jars, and so we end up having to replicate a bit of the `LambdaFunction.kt` class to be able to replace the `APIGatewayProxyRequestEvent` with another one as well.

So if possible, it would be great if the `LambdaFunction` could be adjusted so that we could subclass it, or just wrap it, and be able to pass a custom `request` object (perhaps by allowing the subclass or caller to provide their own equivalent to `asHttp4k`?).

Let me know if you'd be open for that, I'm up to contributing it too.

kiquetal:
Hi @ivoanjo sorry, to bother you, but I really like the use of Lenses, Do you know how can I use http4k within lambdaContext?, I want to implement several fields resolvers for appsync.I understand that the method LambdaFunction::handle receives an APIGatewayProxyRequestEvent so I can test it without an apigateway.

daviddenton:
Just as an FYI, we've recently brought together the implementation mechanisms of the various serverless backend modules to a consistent model. The new way of creating Lambdas with http4k is to subclass LambdaFunction. Details here: https://www.http4k.org/guide/modules/serverless/

ivoanjo:
Sorry for the delay @kiquetal -- did you end up finding a solution to your problem? I'm curious :)

daviddenton:
It might be worth taking another look at the current version of the lambda module. We now support both APIG1/2 and ALB requests, via subclassing with adapters for transforming the request/redponse. I’m not sure if it’s exactly in the right shape but... 🙃

kiquetal:
Hi @ivoanjo , no I couldn't invoke the method without an apigatewayRequest, so I've implemented in nodejs ,unfortunately :(.

ivoanjo:
> It might be worth taking another look at the current version of the lambda module. We now support both APIG1/2 and ALB requests, via subclassing with adapters for transforming the request/redponse. I’m not sure if it’s exactly in the right shape but... upside_down_face

I definitely need to check it out as well :D

aamiel16:
Do we have any example of using lens with the LambdaContext?

daviddenton:
@aamiel16 if you mean RequestContexts then it's the same type of usage as any normal http4k app. To use the RequestContexts in a Lambda - you implement the `AppLoaderWithContexts` and then construct your request context keys as normal as per the examples:

https://www.http4k.org/guide/howto/attach_context_to_a_request/

Issue #292, opened 2019-09-12
Provide structured logging format for DebuggingFilters.*


KushalP:
Structured logging is useful for a [number of reasons](https://stackify.com/what-is-structured-logging-and-why-developers-need-it/). It makes it easier to parse logs using a tool that understands either `key=value` or `json` structures.

Allow configuring the output format for easier integration with tools like [Stackdriver](https://cloud.google.com/logging/docs/structured-logging).

daviddenton:
This is the kind of thing I was going for. Basically we introduce an `EventsFilter` which will allow you to add metadata to an `Event` and finally log it in structured format using an implementation of `Events` which converts the class structure to JSON and dumps it to std out.

Check it out and let me know if you think it would work for you..
https://github.com/http4k/http4k/tree/events

KushalP:
This looks like a good start. It might be useful to understand how you would pass this single (or multiple?) events around as it's hitting different filters and the route in question. Would you `+` multiple events together?

At what point would you serialise them with `slf4j` ?

daviddenton:
Generally, each events instance represents a single "Sink", so for example you'd have one for StandardOut, one for NewRelic, and one for Prometheus metrics, then you can `+` them together into a single "stream" and then use programmatic Filters on each one to determine if you want to consume each Event.

We don't really have much use for logging frameworks like SL4J - most of our stuff fits into the buckets above - in K8S the StandardOut stuff just uses `::println` which is then picked up by FluentD and sent to an aggregator (ELK or Splunk). Of course, there's nothing to stop you implementing one but TBH we find that the very presence of a Logging framework makes people want to vomit random strings instead of structured events. :)

KushalP:
That's interesting because I had issues using http4k by itself without adding slf4j as a dependency. I think I used one of the contract API examples.

This was a grade build error that would appear. I'll try to reproduce it tomorrow when I'm back in the office.

daviddenton:
The main for library doesn't have any dependencies other than kotlin. There may be a transitive dependency on sl4j from some of the modules (you can see it moaning about not being able to find a logger implementation in the travis output from the build) but it shouldn't stop anything building.

daviddenton:
So definitely let us know if you're having trouble. 🙃

KushalP:
I've just checked and the slf4j issue I'm seeing is related to Spek. It has nothing to do with http4k 🤦‍♂ 

My apologies!

daviddenton:
[...]

PR #293, opened 2019-09-13
Moving events around and adding EventsFilters

daviddenton:


KushalP:
This is related to #292

PR #307, opened 2019-11-07
Parse access token response as json when content-type is missing encoding directive

jippeholwerda:
When the `Content-Type` of the response is `application/json` instead of `application/json;charset=utf-8` http4k fails to parse the body as JSON. When no charset is specified, the default is utf-8.

This problem came to light when I was using http4k to authenticate to Keycloak.


daviddenton:
Thanks for noticing this @jippeholwerda! I think we need to actually make the content type equality a bit smarter somehow (as we also use equality in other places in the code, but we can start with this change and iterate.

jippeholwerda:
Yes, I was thinking about that as well but couldn't think of a general solution that is guaranteed to work in all cases. For `application/json` the default charset is `utf-8`, but I'm not sure that will hold for all others.

daviddenton:
This is now released in 3.195.0 :)

jippeholwerda:
I'm already using it :-) Thanks for the quick actions!

PR #310, opened 2019-11-09
Adding to overriding environment should return new overriding environment with added element

jippeholwerda:


daviddenton:
Thanks for another PR!

daviddenton:
3.195.1 is releasing now. :)

jippeholwerda:
Cool, thanks!

Issue #313, opened 2019-11-23
Jackson serialization is not working properly with polymorphic types stored in a collection


alphaho:
Hi, 

I've found that the current `ConfigurableJackson.autoBody` implementation would not work with collections of polymorphic types. It would lose the `@class` field in the serialized json in the below example:

```kotlin
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "@class")
@JsonSubTypes(
    JsonSubTypes.Type(value = FirstChild::class, name = "first"),
    JsonSubTypes.Type(value = SecondChild::class, name = "second")
)
sealed class PolymorphicParent
data class FirstChild(val something: String): PolymorphicParent()
data class SecondChild(val somethingElse: String): PolymorphicParent()

class SampleTest {
    @Test
    fun `roundtrip polymorphic object to and from body`() {
        val body = Body.auto<PolymorphicParent>().toLens()

        val firstChild: PolymorphicParent = FirstChild("hello")
        val secondChild: PolymorphicParent = SecondChild("world")

        assertThat(body(Response(OK).with(body of firstChild)), equalTo(firstChild))
        assertThat(body(Response(OK).with(body of secondChild)), equalTo(secondChild))
    }

    @Test
    fun `roundtrip list of polymorphic objects to and from body`() {
        val body = Body.auto<List<PolymorphicParent>>().toLens()

        val list = listOf(FirstChild("hello"), SecondChild("world"))
        
        assertThat(body(Response(OK).with(body of list)), equalTo(list))
    }
}
```

While the first test(one without being stored in a collection) works, the second test would get:

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Missing type id when trying to resolve subtype of [simple type, class org.http4k.format.PolymorphicParent]: missing type id property '@class'
 at [Source: (String)"[{"something":"hello"},{"somethingElse":"world"}]"; line: 1, column: 22] (through reference chain: java.util.ArrayList[0])
```

daviddenton:
Releasing in 3.199.0 😄

daviddenton:
A quick word of warning before upgrading - using 3.199.0 has caused some breaks in some of my IRL projects. Am investigating what is going on, but I'm seeing these when ```java.lang.reflect.GenericSignatureFormatError: Signature Parse error``` caused by ```	at com.fasterxml.jackson.core.type.TypeReference.<init>(TypeReference.java:33)```

daviddenton:
[...]

PR #314, opened 2019-11-23
Added parameter TypeReference to Jackson writer in ConfigurableJackson.autoBody 

alphaho:
fix issue #313


daviddenton:
Thanks for the PR. :) The change to the reader code seems legit - but I'm slightly worried that I don't understand the change in behaviour introduced in the 3rd commit (fb7be32). 

Is there a test you could write that shows the change in behaviour that would fail with the new version?

alphaho:
Sorry for the confusion. For the 3rd commit (fb7be32), it's just to reuse the extension functions from the library Jackson Kotlin Module(by importing `com.fasterxml.jackson.module.kotlin.convertValue` and `readValue`). 

As I thought it would be a bit more readable to drop the anonymous object creation  `object : TypeReference<T>() {}`. 

These extension functions should not change the behavior as the calls are the same as before after inlined. And the original behavior should already be covered by `JacksonAutoTest`.

I'm happy to drop this commit if it causes too much confusion.

daviddenton:
Ahh! I didn't spot those references. Thanks for clarifying!

alphaho:
No worries, I should have pointed that out in the commit message/PR description.

PR #333, opened 2020-01-08
Add method to allow copying of trace information to child worker threads

jshiell:
This adds method to expose, and allow overwriting, the current Zipkin trace data.

> "Why, this is madness! Why would you break encapsulation in such a way?"

I'm glad you asked. We have a case where we use a thread-pool to query various services in parallel. We need to ensure that these threads have a copy of the current trace data to ensure that we can reconstruct the trace. As such, we need to copy the data across on task initialisation, in a similar manner to the provided test.


daviddenton:
Releasing in 3.208.0

jshiell:
Thanks @daviddenton, much obliged.

PR #334, opened 2020-01-12
Adding token type to access token response, and tidying up response t…

tom:
…o strip out nulls


daviddenton:
Change log? 😉

Tmrw ok?

PR #336, opened 2020-01-13
Introduce support for GZipping response streams

jshiell:
This PR introduces a parameter to the various GZip filters, allowing then to be used with streaming operations, as opposed to the current implementation which relies on in-memory transformations.

This has a number of benefits:
* We can support the GZip en/decoding of resources requiring streaming
* It should reduce memory usage

It also has a number of drawbacks, primarily that you can only consume the stream a single time. As such, it has been introduced as a variation of behaviour, to allow both existing users and convenient applications (e.g. testing) to retain the current behaviour.

And in any case, better devs than I have buggered up streaming code, so it's definitely safer to introduce it as an opt-in option, rather than a mandated replacement.

`GZippingInputStream` is derived from @nicolai-budico's code on [Stack Overflow](https://stackoverflow.com/a/28072468), with a few tweaks (primarily implementing `read(): Int`, to allow its use in pipelined streams, especially for testing).

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/filter/ClientFilters.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/filter/ClientFilters.kt
@@ -136,11 +136,12 @@ object ClientFilters {
     }
 
     /**
-     * Basic GZip and Gunzip support of Request/Response. Does not currently support GZipping streams.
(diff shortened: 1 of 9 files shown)


jshiell:
Cheers @daviddenton, although your confidence scares me shitless :-D

PR #349, opened 2020-01-16
Update index.md

dtanner:
Add suggestion to do your own benchmarks.  
As an example, under real-world loads, our app's performance with Undertow was faster than Apache.  This may or not be true for others.  Another consideration for us was that Undertow took about 10ms to start up in tests, while Apache took about 400ms per test.


daviddenton:
Thanks for the contribution! 🙂

PR #352, opened 2020-01-20
Don't depend on the exception thrown on access to null Java field

pyos:
1.4 changes it to NPE: https://youtrack.jetbrains.com/issue/KT-22275

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
@@ -33,11 +33,7 @@ object JacksonJsonPropertyAnnotated : FieldRetrieval {
         .mapNotNull { f ->
             f.annotations.filterIsInstance<JsonProperty>().find { it.value == name }
                 ?.let { f.name }
-        }.firstOrNull() ?: try {
-        superclass.findName(name)
-    } catch (e: IllegalStateException) {
-        throw NoFieldFound(name, this, e)
-    }
+        }.firstOrNull() ?: superclass?.findName(name)
 }
 
 object JacksonJsonNamingAnnotated : FieldRetrieval {


daviddenton:
Removing the catch of a specific exception wouldn't seem to be the correct thing to do here as it will be changing behaviour. We should either catch both and raise the NoFieldFound or revert to a generic exception to cover the case.

pyos:
This is a private function that's only used a couple lines above the declaration with a `throw NoFieldError` in case of null, so this should be fine.

daviddenton:
Unless I'm being an idiot (I have previous :) ) and am misreading it, 1.4 will now throw an NPE instead of the ISE, which won't result in a null being returned from the method and hence the NFF will be raised? If we want to support both 1.3 and 1.4 then we need to add the extra check for the NPE?

pyos:
To clarify: `superclass` is in fact a call to a Java method `getSuperclass()`. In order to make interoperability easier, Kotlin represents values obtained from Java as so-called "platform types", which have unknown nullability: you can treat them as nullable without incurring warnings about redundant null checks, but also as non-nullable without compile-time errors. However, in the latter case you're effectively asserting that the value is never null, and catching the exception is sonewhat silly (like catching AssertionError) :) The "Impact" sections of the linked issue says so as well.

>hence the NFF will be raised?

Notice that the code is now using the safe subscript operator, `?.` - the value is treated as nullable and there is no assertion.

daviddenton:
We will still need the try/catch to support Kotlin 1.3 and 1.4 - so would this work or do we also need to catch  the NPE that you mentioned above?:

```
try {	      
    superclass?.findName(name)	
} catch (e: IllegalStateException) {	
    throw NoFieldFound(name, this, e)	
}```

daviddenton:
I've making the above change (since it's safe anyway) and releasing it as a part of 3.227.0. But let's clarify if this is enough for 1.4  😄

daviddenton:
I'm going to close this for now - let's reopen if there are further fixes needed..  😄

Issue #340, opened 2020-01-14
Support SameSite cookies


nlochschmidt:
Afaics there is currently no support for setting the `SameSite` attribute for cookies in http4k. 

Would be nice to have a dedicated method for that and probably add the three options `None`, `Strict` and `Lax` as an enum.

I'd even go so far and make `Lax` the default for Cookies as recommended in [1]. Would love to know your opinion on that.

[1]: https://scotthelme.co.uk/csrf-is-really-dead/

danielwellman:
I've made an attempt at adding this support to http4k, see #359.  I made SameSite an enum value but did not set a default if not present because today some browsers inconsistently interpret `SameSite=None`. In those cases it's better to omit the attribute rather than set one; this primarily affects third-party cookies. This means you need to either use two sets of cookies or inspect the request's user agent string and alter the SameSite value of the cookie (not to mention that starting in Chrome 80, all SameSite=None cookies must also be Secure!). See this [web.dev article](https://web.dev/samesite-cookie-recipes/) and the [Chromium list of known incompatible clients](https://www.chromium.org/updates/same-site/incompatible-clients).

Some questions I had on style:
- I made the `SameSite` enum an inner class of `Cookie`, which means clients would need to explicitly import the class if they want to say `Cookie(... sameSite=Lax)` vs. `Cookie(... sameSite=Cookie.SameSite.Lax)`.  What's your preferred way of handling this?
- Parsing an unknown SameSite value (not represented in the enum list) will silently return null, rather than thrown an IllegalArgumentException.  What do you prefer?

I'm still developing my understanding of Kotlin style and am very new to http4k, so please do let me know of any suggestions you have.

daviddenton:
Hey - thanks for the contribution! Generally looks good - I'm not familiar with SameSite so want to read around it a bit.

For your questions:
1. I'd represent SameSite on the top level - we are in the cookie package so that shouldn't be an issue - I can do this after merging it.
2. We don't want to throw in the case of an unknown value - let's start with the null and see how it goes 😄 

Thanks again!

daviddenton:
This will go out in 3.227.0

nlochschmidt:
Thanks!

PR #359, opened 2020-01-26
Support SameSite Cookies #340

danielwellman:
Add the SameSite attribute to Cookies using an enum for the current possible values. SameSite is a modeled here as a nullable attribute without any default value. Although many browsers currently treat a missing SameSite attribute as `Lax` for first-party cookies, this change does not set a default value since some browsers inconsistently handle values of `None` for third-party cookies, notably Safari on Mac OS X 10.14 and iOS 12. To support "SameSite=None" behavior on these browsers, the SameSite attribute should be omitted.

 References:

   - web.dev: SameSite Cookies Explained https://web.dev/samesite-cookies-explained/
   - web.dev: SameSite Cookie Recipes https://web.dev/samesite-cookie-recipes/
   - List of incompatible SameSite=None clients https://www.chromium.org/updates/same-site/incompatible-clients

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/cookie/Cookie.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/cookie/Cookie.kt
@@ -3,6 +3,7 @@ package org.http4k.core.cookie
 import org.http4k.core.Parameters
 import org.http4k.quoted
 import org.http4k.unquoted
+import java.lang.IllegalArgumentException
 import java.time.LocalDateTime
 import java.time.ZoneId
 import java.time.ZonedDateTime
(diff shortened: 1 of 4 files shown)


http4k-core/src/main/kotlin/org/http4k/core/cookie/Cookie.kt
```diff
@@ -15,14 +16,20 @@ data class Cookie(val name: String, val value: String,
...
 
     fun domain(domain: String) = copy(domain = domain)
     fun maxAge(seconds: Long) = copy(maxAge = seconds)
     fun path(path: String) = copy(path = path)
     fun secure() = copy(secure = true)
     fun httpOnly() = copy(httpOnly = true)
     fun expires(date: LocalDateTime): Cookie = copy(expires = date)
+    fun sameSite(sameSite: SameSite) = copy(sameSite = sameSite)
+
+    enum class SameSite {
```
daviddenton: Don't think this shoud be an inner class of the Cookie - let's bump it up to the top level (we are in the "cookie" package :) )

daviddenton: I'll move this after merging it.

PR #374, opened 2020-02-12
add request modifier to OkHttp client

dmgd:


dmgd:
🤦‍♂ can only set timeout on the call, not the request

PR #378, opened 2020-02-20
Eary work on supporting refresh tokens

tom:
Committer: Tom Malone <tomjmalone@gmail.com>


daviddenton:
Do you want to rename the refreshToken package to not have the capital in it?

PR #379, opened 2020-02-21
Apache HTTP Client v5 module

jshiell:
This adds HTTP client support for Apache HTTP Components + Client version 5. Note that the client is still in late beta.

This only covers the synchronous client at present, primarily as the API for the async client has changed a little more and I've run out of time today. Which is, to be frank, Dave's fault.


jshiell:
Now updated to v5 release.

I've only done the sync client, as the async one has significant API changes. Primarily the current invocation has been changed to only support a `SimpleBody`, meaning no streaming of request data. To switch to the producer/consumer API isn't hard, but it's built around NIO and matching it with `InputStream`s is frankly more work than I'm interested in, given I wouldn't be using the async client myself.

daviddenton-teal:
This is now in conflict (my fault - sorry), but QQ - has the API fundamentally changed at all from v4 -> v5 (for the sync client)? 

TBH, not massively keen on having a separate v5 module if we could just use the standard one instead (and just make the async one fall back to the v4 implementation). 

Thoughts?

jshiell:
API for the sync client is pretty consistent - but some of the packges etc have changed. Just enough to be irritating.

PR #385, opened 2020-03-17
Support method not allowed

jshiell:
At present the route-matching is boolean, i.e. a route either matches or it doesn't. This results in a 404 being served for all unmatched requests.

This PR changes the routing so that it returns a match object, allowing us to differentiate on match failures. At this point it represents three states - `matched`, `path matched, method mismatched`, `unmatched`. The result is that we can now support returning 405 where the path is matched but the supported methods differ.

This is a breaking change, as the `Router.match` signature is now `fun match(request: Request): RouterMatchResult`, rather than returning `HttpHandler?`.

Given the scope of this change there's no way I'm merging it without a sanity check 😄

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRoute.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRoute.kt
@@ -10,6 +10,7 @@ import org.http4k.core.NoOp
 import org.http4k.core.Request
 import org.http4k.core.Response
 import org.http4k.core.Status.Companion.BAD_REQUEST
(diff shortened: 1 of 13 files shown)


jshiell:
Poke poke @daviddenton or @s4nchez 😄

daviddenton:
I'm merged this into it's own branch so I can have a good looksee

daviddenton:
BTW @jshiell your import settings are a bit *'y :)

jshiell:
My import settings are the default, life's too short for such shit 😁

daviddenton:
No more PRs from you if you don’t fix it!!

;p

PR #391, opened 2020-04-04
Introduce custom Result (ported from Result4k) and migrated OAuth module to use it

daviddenton:
Want to get away from the direct dependency on Result4k - there is a conversion property that we can use to migrate codebases.


dmcg:
Unsolicited comment I know, but if you are going down this route, I would put the Result types in their own package. I've just noticed that I have files that `import com.natpryce.*` as they pull in Result, Success and Failure, but that risks pulling in anything else that Nat adds to that package.

PR #395, opened 2020-04-16
Add enum StringBiDiMapping

goodhoko:
Hello,
with this PR I try to introduce way to extract/inject enum instances from/to Query and Path.

I have few questions which I will post as comments on the code.

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/lens/query.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/lens/query.kt
@@ -9,3 +9,5 @@ object Query : BiDiLensSpec<Request, String>("query", StringParam,
     LensGet { name, target -> target.queries(name).map { it ?: "" } },
     LensSet { name, values, target -> values.fold(target.removeQuery(name)) { m, next -> m.query(name, next) } }
 )
+
+inline fun <reified T : Enum<T>> Query.enum() = map(StringBiDiMappings.enum<T>())
--- a/http4k-core/src/test/kotlin/org/http4k/lens/BiDiLensSpecTest.kt
+++ b/http4k-core/src/test/kotlin/org/http4k/lens/BiDiLensSpecTest.kt
@@ -4,6 +4,7 @@ import com.natpryce.hamkrest.assertion.assertThat
 import com.natpryce.hamkrest.equalTo
 import com.natpryce.hamkrest.throws
 import org.http4k.base64Encode
+import org.http4k.core.Method
 import org.http4k.core.Uri
 import org.http4k.core.with
 import org.http4k.lens.BiDiLensContract.checkContract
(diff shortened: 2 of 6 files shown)


http4k-core/src/test/kotlin/org/http4k/lens/BiDiLensSpecTest.kt
```diff
@@ -112,6 +113,13 @@ class BiDiLensSpecTest {
     @Test
     fun yearMonth() = checkContract(spec.yearMonth(), YearMonth.of(2000, 2), "2000-02", "", "invalid", "o", "o2000-02", "o2000-022000-02")
 
+    @Test
+    fun enum() {
+        fun BiDiLensSpec<String, String>.enum() = map(StringBiDiMappings.enum<Method>())
```
goodhoko, line 118: Please note, that I declare the `BiDiLensSpec.enum()` method here. The reasons to not have it defined publicly in `lensSpec.kt` like so
```kotlin
inline fun <IN : Any, reified T: Enum<T>> BiDiLensSpec<IN, String>.enum() = map(StringBiDiMappings.enum<T>())
```
are that
1) It would have no usages, because both `Path` and `Query` use `StringBiDiMappings.enum` directly.
2) unfortunately Kotlin is not able to infer the `IN` type when calling the method on the `Query` object. Calling the `enum` extension then becomes impractical
```kotlin
// Request must be specified again
Query.enum<Request, MyEnum>().required("...")
```

I still however wanted to add a test using the `checkContract` helper that's why the test ended up here. In fact it tests functionality from `BiDiMappings.kt` and not `lensSpec.kt` but that's true for most of the tests in this file.

I'm not sure if I took the right decision to play along the general structure of this project.

http4k-core/src/main/kotlin/org/http4k/lens/query.kt
```diff
@@ -8,4 +8,6 @@ typealias QueryLens<T> = Lens<Request, T>
 object Query : BiDiLensSpec<Request, String>("query", StringParam,
     LensGet { name, target -> target.queries(name).map { it ?: "" } },
     LensSet { name, values, target -> values.fold(target.removeQuery(name)) { m, next -> m.query(name, next) } }
-)
\ No newline at end of file
+)
+
+inline fun <reified T : Enum<T>> Query.enum() = map(StringBiDiMappings.enum<T>())
```
goodhoko, line 13: Also note that this commit add new-lines to the end of few files. I noticed that there's a lot of files, that lack NL at EOF even though the `.editorconfig` [mandates otherwise](https://github.com/http4k/http4k/blob/master/.editorconfig#L8). I could submit another PR that fixes this globally. It might also worth it to setup some linter eg. [ktlint](https://ktlint.github.io/) to have generally consistent formatting.

daviddenton:
LGTM! :)

PR #414, opened 2020-05-11
Fix BasicAuth server filter to not throw an exception on invalid base64 input

Sebruck:
Currently this filter throws an exception which results in an internal server error when the given base64 string is invalid (cut of byte sequence, invalid characters...).

This is bad since it's not a server but a client error. After this change, it will lead to an UNAUTHORIZED when a client submits an invalid base64 string.


daviddenton:
good catch! 🙃

Sebruck:
That was quick, thx for merging!

daviddenton:
Releasing now in 3.247.0. thanks again. :)

PR #415, opened 2020-05-12
Add support for securing the API description endpoint

goodhoko:
In our use case we needed to have the swagger JSON under authorised access. I wasn't able achieve that with the current public http4k APIs. Hence this MR.

Peace! .)


daviddenton:
Will merge this, but for the moment you could compose a presecurity filter with a custom filter to check for the API route and add security there. :)

goodhoko:
@daviddenton Thanks!

Also thanks for the eyeopening (and obivous 🙈) solution.

PR #429, opened 2020-05-30
Initial implementation of the adapter for the Google Cloud Functions

ssijak:
Google Cloud Functions can now run java11 runtime.

I have Implemented a simple adapter for their interface which is invoked when the function is run.

Tested it both locally and in the cloud and it works fine.

If you wan't to be able to invoke them locally this is the googles suggested way for Gradle  (need to pass -PrunFunction.target parameter which is the <package>.<MainClassName> which implements the interface)

```
configurations {
    invoker
}

dependencies {
    implementation 'com.google.cloud.functions:functions-framework-api:1.0.1'
    invoker 'com.google.cloud.functions.invoker:java-function-invoker:1.0.0-alpha-2-rc5'
}

tasks.register("runFunction", JavaExec) {
    main = 'com.google.cloud.functions.invoker.runner.Invoker'
    classpath(configurations.invoker)
    inputs.files(configurations.runtimeClasspath, sourceSets.main.output)
    args(
            '--target', project.findProperty('runFunction.target'),
            '--port', project.findProperty('runFunction.port') ?: 8080
    )
    doFirst {
        args('--classpath', files(configurations.runtimeClasspath, sourceSets.main.output).asPath)
    }
}
```

Testing or deploying to the cloud from gradle must be done by first building the JAR with for example [this plugin](https://github.com/johnrengelman/shadow) and than deploying the JAR with `gcloud` CLI. For example with 
```
gcloud functions deploy example-function \
    --runtime=java11 \
    --entry-point=org.http4k.serverless.gcp.GCFunctionExample \
    --trigger-http \
    --source=libs/
```

Example function implementing the adapter : 
```
package org.http4k.serverless.gcp

import org.http4k.core.*
import org.http4k.core.Status.Companion.INTERNAL_SERVER_ERROR
import org.http4k.core.Status.Companion.OK
import org.http4k.routing.bind
import org.http4k.routing.path
import org.http4k.routing.routes

val r = routes(
    "/hello" bind routes(
        "/{name:.*}" bind Method.GET to { request: Request -> Response(OK).body("Hello, ${request.path("name")}!") }
    ),
    "/gcf" bind Method.GET to { request: Request -> Response(OK).body("works!") }
    "/fail" bind Method.GET to { request: Request -> Response(INTERNAL_SERVER_ERROR) }
)
val timer = Filter { next: HttpHandler ->
    { request: Request ->
        val start = System.currentTimeMillis()
        val response = next(request)
        val latency = System.currentTimeMillis() - start
        println("I took $latency ms")
        response
    }
}

class GCFunctionExample : Http4kGCFAdapter(timer.then(r))
```

Being my first pull request here, comments and requests for improvements are wellcome :) I can also add the documentation when/if this is merged.


knyttl:
Would you maybe include your comment above as README?

knyttl:
I tried to recreate your example and I keep getting:

Screenshot 2020-05-30 at 18 04 29

The same when running with gradle. Not sure how and why this happens.

ssijak:
@knyttl Yes, I added `functions-framework-api` as implementation in the `http4k-serverless-gcp` , now I changed it to compile so that users of the lib would not need to add the dependency.

daviddenton:
Thanks for the awesome contribution! Will merge this in now.

In the near future, we'll see if there is some commonality between the Lambda solution and this module that can be abstracted or reused to provide a consistent experience across both platforms (like we do with the Server backend modules).

But once again.. thank you! 😄 😄

daviddenton:
We are also going to rename the module to either `http4k-serverless-gcf` or maybe `http4k-serverless-googlecloudfunctions` (which is a mouthful, but.. 🤷 ) 😄

knyttl:
I was playing a bit with it and I hope someone could disprove my observations, because I am not very happy with them.

1. Running http4k in Docker container in GCloud Run starts up in ~10 seconds and it will most probably stay alive for long time – that is not too bad.
2. Running http4k as CF starts up in ~40 seconds (!) and it will most probably be killed very soon, so the 40s start up will have to be waited through again soon. Which is not cool :/

daviddenton:
I've not played with GCF personally, but the overhead introduced by using the http4k function wrapper should be absolutely minimal. Even without proguarding (to reduce JAR size), the core http4k JAR is <1mb.

With Lambda and Proguard, we'd managed to get a simple helloworld function down to a few hundred k of size.

If we just do a barebones helloworld with no http4k, how does that perform?

daviddenton:
I've now made the API of the module consistent with the AWS Lambda version - you now just need to implement an instance of the `AppLoader` interface which provides the HttpHandler from the environment map - instead of passing the handler directly. It's a slight bit of an extra step but it's TBH consistency for all Serverless implementations is easier from a user point of view.

Let's get an initial version out there in 3.249.0 and we can see if we can get to the bottom of the performance thing. I also want to get Azure and OpenFaaS implementations done with the same interface.. :)

PR #430, opened 2020-05-30
Add response streaming to Netty server

carbotaniuman:
I'm unsure of what this line does, but it's just been replaced by `addListener(CLOSE)`. This currently implements response streaming but not request streaming, as request streaming likely requires a complete rewrite.
```
if (request.decoderResult() == SUCCESS) addListener(CLOSE) else ctx.close()	
```


s4nchez:
@carbotaniuman thank you for the PR!

PR #434, opened 2020-06-04
Implementation of remote address and port in Request interface

### LINKED ISSUE #120: No equivalent to `ServletRequest.getRemoteAddr()`
There currently isn't any way to get the "remote address", a la  `ServletRequest.getRemoteAddr()`.

As @daviddenton mentioned on slack, "you need to be a bit generally wary of outside reported IPs as they can get modified by every proxy server that they go through (and not always in a correct way!)". However, they are sometimes useful for debugging.

For example, I was trying to make an http4k `Filter` equivalent to [`AWSXRayServletFilter`](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/javax/servlet/AWSXRayServletFilter.java), and it stores the remote address in the trace segments.

I investigated the existing server implementations, and found that they all have a way of getting this information. I've recorded what I found below, but keep in mind that none of these code snippets are actually tested.

    // Undertow
    InetAddress addr = httpServerExchange.getSourceAddress();

    // Apache
    HttpInetConnection connection =
        (HttpInetConnection) httpContext.getAttribute(ExecutionContext.HTTP_CONNECTION);
    InetAddress addr = connection.getRemoteAddress();

    // Jetty
    InetAddress addr = InetAddress.getByName(httpServletRequest.getRemoteAddr());

    // Netty
    InetAddress addr =
        ((InetSocketAddress) (channelHandlerContext.channel().remoteAddress())).getAddress();

kam1sh:
closes #120

Hi! Thank you for http4k. I really like it.

I implemented support for source address and port in Request interface.
Added two tests for them, but servers have been tested manually (but I can try to write junit tests, if there's need in them).

A few moments:

* What will be default values, when address and port isn't specified? "" and -1 or null/null?
* Ratpack: for IPv6 loopback source address is 0:0:0:0:0:0:0:1%0. Just for your information.
* KtorNetty/KtorCIO: I couldn't find any information about retrieving source port.

Also. I'm getting errors in some tests and I don't understand how I could affect them, for example, in `src/docs/tutorials/tdding_http4k/_1/tests.kt` (IllegalStateException: server.uri must not be null)

Looks like the root cause is in this line:

https://github.com/http4k/http4k/blob/e09e2f6db7a66899728339967c7df25f0c62a76b/http4k-server-jetty/src/main/kotlin/org/http4k/server/jetty.kt#L38

Do you know what this `else server.uri.port` means?

Also 2: I kinda downgraded Gradle because `./gradle wrapper` was updating to a newer version that is incompatible with cobertura plugin.

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -116,8 +116,9 @@ object Versions {
      * Current version: "6.1.1"
      * See issue 19: How to update Gradle itself?
      * https://github.com/jmfayard/buildSrcVersions/issues/19
+     * Why 6.1.1? See https://github.com/stevesaliman/gradle-cobertura-plugin/issues/168 for details.
      */
-    const val gradleLatestVersion: String = "6.4.1"
+    const val gradleLatestVersion: String = "6.1.1"
 }
 
 /**
--- a/http4k-core/src/main/kotlin/org/http4k/core/http.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/http.kt
@@ -7,6 +7,7 @@ import org.http4k.lens.WebForm
 import org.http4k.routing.RoutedRequest
 import java.io.Closeable
 import java.io.InputStream
+import java.net.InetAddress
 import java.nio.ByteBuffer
 
 typealias Headers = Parameters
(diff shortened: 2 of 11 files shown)


s4nchez:
Thank you for your contribution, @kam1sh!

s4nchez:
@kam1sh I've mistakenly merged that. Sorry about that. We're still reviewing the best way to introduce that change and will let you know as soon as we have a clear view on the API.

Sorry for the confusion!

kam1sh:
@s4nchez No problem =) 

> will let you know as soon as we have a clear view on the API

OK. I'll wait with interest! If what, I'm still subscribed to the issue (120).

PR #435, opened 2020-06-07
Add Facebook OAuth provider

knyttl:
I wanted to add also an iCloud provider but that seems to be not that easy :/

### DIFF
--- a/http4k-security-oauth/src/main/kotlin/org/http4k/security/oauthProviders.kt
+++ b/http4k-security-oauth/src/main/kotlin/org/http4k/security/oauthProviders.kt
@@ -31,6 +31,15 @@ fun OAuthProvider.Companion.dropbox(client: HttpHandler, credentials: Credential
         listOf(""),
         oAuthPersistence)
 
+fun OAuthProvider.Companion.facebook(client: HttpHandler, credentials: Credentials, callbackUri: Uri, oAuthPersistence: OAuthPersistence, scopes: List<String> = listOf("email")): OAuthProvider =
+    OAuthProvider(
+        OAuthProviderConfig(Uri.of("https://www.facebook.com"), "/dialog/oauth", "/oauth/access_token", credentials, Uri.of("https://graph.facebook.com")),
+        client,
+        callbackUri,
+        scopes,
+        oAuthPersistence
+    )
+
 fun OAuthProvider.Companion.google(client: HttpHandler, credentials: Credentials, callbackUri: Uri, oAuthPersistence: OAuthPersistence, scopes: List<String> = listOf("openid")): OAuthProvider =
     OAuthProvider(
         OAuthProviderConfig(Uri.of("https://accounts.google.com"), "/o/oauth2/v2/auth", "/oauth2/v4/token", credentials, Uri.of("https://www.googleapis.com")),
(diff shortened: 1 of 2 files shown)


s4nchez:
Thank you for your contribution, @knyttl!

daviddenton:
@knyttl SIWA (appleId) should be pretty easy to add, apart from the fact that we haven't fully implemented form_post response mode :) . In fact, the enum value for it *is* there but it's commented out, so I think that all the machinery should be there to make it work. I'm not an expert though, so I could be wrong! :)

knyttl:
I think I will have a PR for apple tonight, but I am not very happy about
it. The impl on apples side is not ideal.

daviddenton:
It looks like we need to implement this type of mechanic, which isn't very nice - not sure if it fits entirely with the current model:

https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html#FormPostResponseExample

knyttl:
Hm, there is something behind CSRF. I tried following:

```
fun OAuthProvider.Companion.apple(client: HttpHandler, credentials: Credentials, callbackUri: Uri, oAuthPersistence: OAuthPersistence, scopes: List<String> = listOf("openid email")): OAuthProvider =
    OAuthProvider(
        OAuthProviderConfig(Uri.of("https://appleid.apple.com"), "/auth/authorize", "/auth/token", credentials, Uri.of("https://appleid.apple.com")),
        client,
        callbackUri,
        scopes,
        oAuthPersistence
    )
```

It works until the callback gets back to my app:

http://localhost:8080/apple/v1/callback?state=csrf=CSRF-CODE&uri=AUTH_URI&code=XXX.YYY

Which return 403. I tried to debug and got here:

```
class OAuthCallback(
    private val oAuthPersistence: OAuthPersistence,
    private val idTokenConsumer: IdTokenConsumer,
    private val accessTokenFetcher: AccessTokenFetcher
) : HttpHandler {

    override fun invoke(request: Request) = request.queryOrFragmentParameter("code")
        ?.let { code ->
            val state = request.queryOrFragmentParameter("state")?.toParameters() ?: emptyList()
            state.find { it.first == "csrf" }?.second
                ?.let(::CrossSiteRequestForgeryToken)
                ?.takeIf { it == oAuthPersistence.retrieveCsrf(request) }
                ?.let {
```

Here the state contains `csrf` key, but its value is `null`. Is the URL somehow malformed?

tom:
Hmm I'm a little confused, I had a quick look at the docs last night, I thought that Apple had a mandatory response_mode of form_post, would not have caused the callback be a post not a get? That would then make sense as to why the state is not being picked up as its a form post, or does Apple support a query response_mode?

Also what OAuthPersistence are you using, when you say state is null is that from the request or OAuthPersistence?

I'll try and have a look at this tomorrow

knyttl:
> Hmm I'm a little confused, I had a quick look at the docs last night, I thought that Apple had a mandatory response_mode of form_post, would not have caused the callback be a post not a get?

It seems it sends a GET unless you require additional scopes. Once you require different scopes, it will state something like: "response_mode=post" is required. – So that is another issue.

Back to this issue, It seems to me there is some encoding issue as the second `=` within `state=csrf=CSRF_CODE` is not encoded. I checked and it is encoded within google or facebook auth dance.

PR #439, opened 2020-06-09
Expose client address and port in Request

### LINKED ISSUE #120: No equivalent to `ServletRequest.getRemoteAddr()`

s4nchez:
Closes #120

Based on #434. Simplified the API in `Request` and added tests for it in `ServerContract`.

It's worth noticing that this capability is inherently limited:
1. It depends on what the server implementation retrieves from the client socket. While most servers will bring an IP address, Ktor returns the client hostname.
2. It has very limited use if the server sits behind any proxy that masks client connections.

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -116,8 +116,9 @@ object Versions {
      * Current version: "6.1.1"
      * See issue 19: How to update Gradle itself?
      * https://github.com/jmfayard/buildSrcVersions/issues/19
+     * Why 6.1.1? See https://github.com/stevesaliman/gradle-cobertura-plugin/issues/168 for details.
      */
-    const val gradleLatestVersion: String = "6.4.1"
+    const val gradleLatestVersion: String = "6.1.1"
 }
 
 /**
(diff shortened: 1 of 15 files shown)


s4nchez:
H/T to @kam1sh  for the initial implementation this is based on...

PR #442, opened 2020-06-11
Add request scheme to RequestSource, where available

jshiell:
This PR adds a `scheme` property to the `RequestSource` object, allowing the inference of HTTPS use where HTTPS is handled directly on the server. Do note that not all servers expose this (e.g. Netty), therefore it's only supported by a subset of servers, returning `null` if not supported.

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/http.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/http.kt
@@ -329,7 +329,7 @@ data class MemoryResponse(override val status: Status, override val headers: Hea
         && body == other.body)
 }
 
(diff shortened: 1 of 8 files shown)


jshiell:
Not sure what's going on with the checks - `WebsocketClientTest` seems to be all over the place, working in some builds, failing in orgers (e.g. Shippable). I can't get it to reproduce locally.

s4nchez:
Yes, the websocket tests have become flaky recently :( We need to get some time to investigate which change introduced the instability...

daviddenton:
Am pretty sure it was the reconnection test change

jshiell:
Just as long as it wasn't me 😅

PR #447, opened 2020-06-15
Added support for multiple "cookie" headers

jshiell:
HTTP/2 [allows multiple `Cookie` headers](https://tools.ietf.org/html/rfc7540#section-8.1.2.5). This changes the `cookies()` extension method to deal with this, rather than just grabbing the first one.


daviddenton:
Looks good. Ship it!

jshiell:
Cheers, I was waiting for codebeat to finish. But Christ it’s takin its time.

PR #450, opened 2020-06-22
Add Apache HTTP Components Client & Server for version 5

jshiell:
This replaces the default Apache client & server modules with version 5 of the client & core.

The version 4 client/server is retained in the `*-apache4` modules.

Note that the v5 async client only supports in-memory bodies at present, due to changes in the API. Streaming support with take a bit more reworking, if desired.


jshiell:
Damned if I know why Travis is failing; it's the JaCoCo report, but I can't find the root cause.

daviddenton:
Jacoco does this if there are any empty source directories. Try rebasing from master?

jshiell:
It's based off 3f166dabb, which seems to be current master.

I've manually checked the modified modules; can't see any empty source dirs I fear.

jshiell:
Oh FFS, Jacoco doesn't handle module separation, and the classes are called the same thing.

```
Can't add different class with same name: org/http4k/client/PreCannedApacheHttpClients$insecureApacheHttpClient$1
```

Guess I'll have to rename the classes.

jshiell:
Damned if I know what to call them though without either breaking the easy module swap for v4, or distinguishing v5 from the standard names.

daviddenton:
Yes, it’s crap isn’t it.. :)

jshiell:
Done, renaming the v4 classes. Which makes migration very slightly more complicated, but keeps the new stuff clean. Death to JaCoCo!

daviddenton:
That stuff is all pretty new anyway- it’s possible you could do something with top level functions to keep it compatible?

jshiell:
It seems to be anything that conflicts in the namespacing, so unless you're going to have a wrapper object that delegates to both dynamically, I've no bright ideas on how to do so.

jshiell:
Thanks Ivan 👍

daviddenton:
FYI, we will need to get the new packages added to JCenter before anyone can use them. I have also temporarily disabled the maven sync scripts to the new modules until that's done.

jshiell:
Cool cool. In no desperate rush for this one in any case, just planning ahead.

PR #453, opened 2020-07-01
add composite (pageable) example

pabl0rg:


daviddenton:
I've added this into the docs now, so we can close this. Thanks for the idea though @pabl0rg  😄

Issue #427, opened 2020-05-26
Default ServerConfig implementations do not shut down gracefully


daviddenton:
At the moment, several of the server backends will terminate in-progress connections instead of shutting down gracefully. Wherever possible, we should make the default to do a graceful shutdown - this might not be available given the underlying server APIs available in each.

Need to write a test in `ServerContract` and handle each backend in turn to see if we can make it pass.

nlochschmidt:
I think we are seeing this with the SunHttp server when using http4k with OpenFaaS. 

@daviddenton you seem to think that there are backends that already do graceful shutdown? Do you happen to know which ones?

daviddenton:
undertow has a graceful shutdown handler that you can use. we should really put that into the default config.

also, beware startup time - sunhttp is especially good for this.

I'd also like to submit an official template for http4k on openfaas but haven't got around to it.

daviddenton:
I was actually surprised that openfaas uses the sun server. it's not really meant for prod use AFAIK... we have had concurrency issues with it in the past. not sure what the openfaas lifecycle model.is or if it routes multiple requests to a single instance at once.

nlochschmidt:
OpenFaaS is not using it, it's just that we are :wink:. We had an exchange around that on Twitter and @alexellis mentioned that he's using the same one for the Java templates

https://twitter.com/http4k/status/1161552909051604992?s=20

nlochschmidt:
I don't want to take this off-topic so: 

As far as I understand it the Sun HttpServer allows us to specify a timeout on `stop` in seconds, however it is [currently hardcoded to `0`]( https://github.com/http4k/http4k/blob/master/http4k-core/src/main/kotlin/org/http4k/server/SunHttp.kt#L32). Would the idea be to have a new method e.g. `gracefulStop(timeout: Duration)` or just `stop(timeout: Duration)` and adapt that to each server implementation? Or would you rather see this as part of the server configuration?

daviddenton:
I'd ideally like to see all of the server's do graceful shutdown by default (unless it makes them slow to stop if there are no requests). Each of the ServerConfigs should not take any extra config (to make things simple), but also be structured in a way that people can implement their own versions with minimum effort - eg. reusing the Servlet in the Jetty, or the handler in Undertow.

nlochschmidt:
[...]

PR #462, opened 2020-08-02
Graceful shutdown on server stop

nlochschmidt:
PR for #427 

Implemented using a `GracefulShutdown`-Filter (based on Undertows builtin handler) that allows to initiate a shutdown on a handler (Handler returns 503). The shutdown call on the filter is blocking until all requests have successfully been handled at which point we can safely shutdown the server.

The difficult part here is that some of the servers take a long time to start and/or shutdown too early which makes the tests difficult to write.

Todos
- [x] SunHttp
- [ ] ApacheServer
- [ ] Apache4Server
- [ ] KtorCIO
- [ ] KtorNetty
- [x] Ratpack
- [ ] Jetty
- [ ] Undertow
- [x] Netty

- [x] Tests for the `GracefulShutdown`-Filter


nlochschmidt:
@daviddenton @s4nchez This turns out to be quite tricky. I thought I had a working solution and then I made my slow-echo function use some slow streaming for sending the body and the majority of servers failed the test again.

At this point I would like to get feedback on the general approach and the test cases I have build so far.

The general idea that I am following is to make the shutdown behavior of all server implementations consistent. My idea is that when stop is called, the server immediately starts responding with `503 SERVICE UNAVAILABLE` to all incoming requests.  In a production environment, this would typically mean that the service will be taken out of the load balancer and many http clients will simply retry and hopefully end up on a different instance. Once `stop()` returned it should no longer be possible to contact the server.

Since not all the servers provide this kind of functionality, I decided to implement it as a new `ServerFilter` (`GracefulShutdown`).

Please have a look at the changes I made to ServerContract
https://github.com/http4k/http4k/pull/462/files#diff-0ce83423b8e07c3180a20547f64f4cde

and how I implemented the graceful shutdown filter
https://github.com/http4k/http4k/pull/462/files#diff-867cbc1dc9997f715ab980f330e7c08f

I hope this is going into the right direction.

http4k-core/src/test/kotlin/org/http4k/server/ServerContract.kt
```diff
@@ -65,12 +79,19 @@ abstract class ServerContract(private val serverConfig: (Int) -> ServerConfig, p
...
+                Response(OK).with(Body.binary(TEXT_PLAIN).toLens() of SlowBodyProducer(it.bodyString()))
             }
         )
 
     @BeforeEach
     fun before() {
         server = routes(*routes.toTypedArray()).asServer(serverConfig(0)).start()
+        retry(seconds=3) {
+            assertThat(client(Request(GET, "$baseUrl/health")), hasStatus(OK).and(hasBody("UP")))
+        }
```
nlochschmidt, line 94: This resolved some test flakiness when servers where slow to start.

nlochschmidt:
Closing this one in favor of the simpler approach in #469

PR #466, opened 2020-08-05
Change defaults of SunHttp server and enable reuse of the HttpExchangeHandler

### LINKED ISSUE #465: Improve the defaults of SunHttp server
While working on #462 I noticed that the SunHttp server has very naive settings. I do realize that you might not want developers to actually use that server in production and instead opt to keep it as it is, but maybe at least some of what I would propose still makes sense:

- Turn the private extension functions `HttpExchange.populate(...)` and `HttpExchange.toRequest()` into a public `SunHttpRequestHandler` to make them reusable for customised SunHttp instances
- Set a default backlog for the ServerSocket (e.g. `1000` as in ApacheServer, Undertow and Netty)
- Set an executor service to allow parallel request handling (e.g. based on available processors as in Undertow and Netty)

What do you think?

nlochschmidt:
Included in this MR:

- **Refactoring to allow reuse of the handler** 
Done in a very similar way to the Undertow handler and fixed a potential issue with HttpExchange not getting closed in the rare circumstance that `sendResponseHeaders(500, 0)` inside the `catch` block fails.
- **Enable parallel processing of requests**
Opted for using a work-stealing pool aka (ForkJoinPool) because it doesn't require us to set a fixed limit of threads and instead allows us to set the targeted parallelism which by default is `getRuntime().availableProcessors()`.
- **Set backlog of the server socket to 1000**
This is the same limit that is applied explicitly or through the defaults in the case of Netty, Undertow and ApacheServer

Closes #465

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/server/SunHttp.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/server/SunHttp.kt
@@ -10,7 +10,9 @@ import org.http4k.core.RequestSource
 import org.http4k.core.Response
 import org.http4k.core.Uri
 import org.http4k.core.safeLong
+import java.lang.Runtime.getRuntime
 import java.net.InetSocketAddress
+import java.util.concurrent.Executors.newWorkStealingPool
 
 class HttpExchangeHandler(private val handler: HttpHandler): SunHttpHandler {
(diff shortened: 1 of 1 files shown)


s4nchez:
Thank you for the PR, @nlochschmidt !

PR #469, opened 2020-08-10
Graceful server shutdown

### LINKED ISSUE #427: Default ServerConfig implementations do not shut down gracefully

nlochschmidt:
This MR makes all server implementations perform graceful shutdown on `stop()`. As before, `stop()` will block until the server has completed the shutdown.

Compared to a previous MR #462 this MR does not make the actual shutdown behavior consistent. Some implementations will start returning 503 during shutdown (Jetty and Undertow) while the others will close the server socket immediately.

Most servers will also return from `stop()`  sooner in case there are no more inflight requests. However, servers handle idle connections (keep-alive) differently on shutdown. ApacheServer (4 and 5) as well as NettyKtor will not return early and will keep the keep-alive connections open until the very end. Because of this, I decided to set the default shutdown timeout lower for all the implementations (3 seconds currently).

Due to the above, the tests for ApacheServer, Apache4Server, KtorNetty will now take significantly longer (around 40sec) because the server is stopped after each test.

FYI I asked about the behavior of the ApacheServer [on their mailing list](http://mail-archives.apache.org/mod_mbox/hc-httpclient-users/202008.mbox/%3C336d98ce93e13ad939797c3f2acaab35d3652d83.camel%40apache.org%3E)

Closes #427


knyttl:
I think I got into the issue which this MR tries to fix – when running http4k in Google Cloud Run, it seems to be running (and logging) some time after it should already be stopped.

I would be up for merging this. 👍

s4nchez:
@knyttl this PR seems very out of date and at the time we've talked to @nlochschmidt about how we could approach this in a server-independent way.

That means we'd represent three modes to shutdown a server:
* Immediate (breaking any ongoing request)
* Time-based (give some time for ongoing requests to finish and then shutdown)
* Gracefully (don't shutdown until all ongoing requests are finished)

We expect different backends to support only a subset of those modes and would require the server startup to fail fast if they are configured in an unsupported way.

@nlochschmidt IIRC you had made a bit more progress on your fork based on that conversation. Do you still have that so we can try to progress this feature?

nlochschmidt:
@s4nchez thanks for reminding me about this. The experimental code with the different stop modes is on this branch https://github.com/http4k/http4k/compare/master...nlochschmidt:427-graceful-stop-modes. I sent you the branch on the Kotlin Slack and was waiting for feedback from you, but then also totally forgot about it too 🙈.

So do you think I should discard the current commits in this PR in favor of the stop mode approach? I could then attempt to get the remaining servers working as well.

_Btw: This hasn't been that important for me/us because we are using a custom http server implementation that does not adhere to the existing server contract._

s4nchez:
@nlochschmidt Your new approach looks definitely more promising than this PR, so happy to continue from there. The only concern I have about that at the moment is testing. If it’s too slow or brittle we may need to run it differently...

http4k-server/netty/src/test/kotlin/org/http4k/server/NettyTest.kt
```diff
@@ -2,4 +2,11 @@ package org.http4k.server
...
 
-class NettyTest : ServerContract(::Netty, ApacheClient())
\ No newline at end of file
+class NettyTest : ServerContract(::Netty, ApacheClient())
+
+class NettyStopTest : ServerStopContract(
+    { stopMode -> Netty(0, stopMode) },
+    ApacheClient(),
+    {
+        enableGracefulStop()
```
nlochschmidt, line 11: I didn't manage to implement `Immediate` stop mode for Netty.

http4k-server/ktorcio/src/test/kotlin/org/http4k/server/KtorCIOTest.kt
```diff
@@ -27,3 +27,12 @@ class KtorCIOTest : ServerContract({ KtorCIO(Random().nextInt(1000) + 8745) }, A
...
     override fun requestScheme(): Matcher<String?> = equalTo("http")
 }
+
+class KtorCIOStopTest : ServerStopContract(
+    { stopMode -> KtorCIO(Random().nextInt(1000) + 8745, stopMode) },
+    ApacheClient(),
+    {
+        enableImmediateStop()
+        enableGracefulStop()
+        enableDelayedStop()
```
nlochschmidt, line 37: Not sure if `Delayed` stop should be added when `Graceful` already works as `Delayed` should only be for cases where `Graceful` is not possible

http4k-server/ratpack/src/test/kotlin/org/http4k/server/RatpackTest.kt
```diff
@@ -6,4 +6,6 @@ class RatpackTest : ServerContract({ Ratpack(it) }, ApacheClient()) {
 
     override fun `ok when length already set`() {
     }
-}
\ No newline at end of file
+}
+
+class RatpackStopTest : ServerStopContract({ Ratpack(0, it) }, ApacheClient(), { enableGracefulStop() })
```
nlochschmidt, line 11: I don't think it's possible to have `Ratpack` shutdown immediately.

nlochschmidt:
I updated the branch and pushed the remaining server shutdown implementations. I tried to set the stop mode defaults for the various servers so that they replicate the current behavior.

There are two known issues that I couldn't resolve:

1. Tests don't run when my firewall is on because the requests are made agains `0.0.0.0` instaead of `localhost` or `127.0.0.1`. Switching to the loopback addresses causes the ApacheServer test to fail.
2. The the KtorCIO test is kind of flaky and is sometimes closing the channel to early resulting in `Premature end of chunk coded message body: closing chunk expected` and a failed test.

@s4nchez @daviddenton I won't be able to continue working on this until next week, so if you could put on the finishing touches, I would appreciate that a lot.

s4nchez:
Hey, @nlochschmidt sorry for not looking into this yet. Life's got in the way. I'll try to pick it up on the next few days.

pabl0rg:
this feature would be very useful

s4nchez:
@nlochschmidt, I finally had a chance to work a bit on this:

I've removed the Delayed option as most servers won't provide that option out-of-the-box. 

I've also moved the tests to a new module `shutdown-integration-tests` and changed them to work by running servers inside Docker. Stopping containers is more reliable and consistent with the usual use-case of the feature. Those tests won't run in the standard build as they're slow/brittle, but can be run via `./gradlew :http4k-server-shutdown-integration-test:integrationTests`

Graceful shutdown is available from version `4.25.13.0`. Thank you again for working with us on this one :)

nlochschmidt:
Thanks a lot @s4nchez for picking this up again.

Since we have been able to simply use our own server implementation, this wasn't blocking us, but it is really good have it built right into http4k 👍

PR #488, opened 2020-09-19
Added originRegex to CorsPolicy

### LINKED ISSUE #487: Regex input for allowing origins in CorsPolicy
The current logic for verification of the CORS policy only checks if the list of `origins` of `CorsPolicy` contains the origin of the request.
```
val allowedOrigin = when {
                    "*" in policy.origins -> "*"
                    origin != null && origin in policy.origins -> origin
                    else -> "null"
                }
```

Would it be okay to have a regex input for the `CorsPolicy` such that any origins matching the regex are allowed?

For eg, if I want domains like `aaa.company.com`, `bbb.company.com` and `ccc.company.com`, we could input a single regex to match all the 3, rather than have a list of those 3 domains in the CorsPolicy, which wouldn't scale when I have to add any more similar domains.

kratostaine:
This change is to provide a Regex input for allowing origins in `CorsPolicy`. Closes #487

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/filter/ServerFilters.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/filter/ServerFilters.kt
@@ -34,7 +34,8 @@ import java.io.StringWriter
 data class CorsPolicy(val origins: List<String>,
                       val headers: List<String>,
                       val methods: List<Method>,
-                      val credentials: Boolean = false) {
+                      val credentials: Boolean = false,
+                      val originRegex: Regex = Regex("")) {
     companion object {
         val UnsafeGlobalPermissive = CorsPolicy(listOf("*"), listOf("content-type"), Method.values().toList(), true)
     }
@@ -55,7 +56,7 @@ object ServerFilters {
                 val origin = it.header("Origin")
                 val allowedOrigin = when {
                     "*" in policy.origins -> "*"
-                    origin != null && origin in policy.origins -> origin
+                    origin != null && (policy.originRegex.matches(origin) || origin in policy.origins) -> origin
                     else -> "null"
                 }
 
--- a/http4k-core/src/test/kotlin/org/http4k/filter/ServerFiltersTest.kt
+++ b/http4k-core/src/test/kotlin/org/http4k/filter/ServerFiltersTest.kt
@@ -136,6 +136,30 @@ class ServerFiltersTest {
     }
 
     @Test
+    fun `OPTIONS - requests are returned with expected headers when cors policy has origin pattern`() {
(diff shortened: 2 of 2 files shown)


kratostaine:
@daviddenton There's this `org.http4k.chaos.LatencyBehaviourTest > latency behaviour should add extra latency()` test that's failing in Travis, but not sure why, as it's passing consistently on my machine

kratostaine:
@daviddenton your thoughts on this please?

daviddenton:
Am actually on hols at the moment... back next week so will take a proper look then. 🙃

kratostaine:
Ah, didn't realize that 😁 Apologies for the disturbance! Have a great holiday 😃

daviddenton:
No problem. 🙃

daviddenton:
So have had a quick look - I think that there is a little more we can do here to remove the duplication between the list of origins and the regex. If we introduce `interface OriginPolicy : (String?) -> Boolean` or the equivalent typealias, we can then provide a couple of implementations which will handle any case. eg: 

1. single origin
2. list of origins
3. regex on origin

It's a bit more work, but I think it will provide a better API experience.

kratostaine:
@daviddenton Do you mean to say, create 3 separate policy implementations for single origin, list of origins and origin regex by inheriting from the originPolicy interface, moving the origin matching to the respective implementations and then having the logic in filter as
```
val allowedOrigin = if (policy.originPolicy(origin)) {
                    origin
                } else {
                    "null"
                }
```

daviddenton:
Along those lines, yes. 

For the break - we can work around it. We can also add another overloaded method which takes the list and just defaults to using the correct policy - then deprecate the old method. 🙃

kratostaine:
@daviddenton you mean create a new class similar to `CorsPolicy`, replacing the `origins: List<String>` param with `originPolicy: OriginPolicy`, and have an overloaded method with the new class as input param right? Just trying to ensure that I'm aligned to your thoughts.

kratostaine:
@daviddenton you mean create a new class similar to `CorsPolicy`, replacing the `origins: List<String>` param with `originPolicy: OriginPolicy`, and have an overloaded method with the new class as input param right? Just trying to ensure that I'm aligned to your thoughts.

daviddenton:
Yep. That’s about the size of it. 🙃

daviddenton:
Not exactly. Just change CorsPolicy to always take the OriginPolicy and then overload the constructor to take the list of origins and wrap it in whichever implementation takes the list.

kratostaine:
Sounds good. Thank you, David. I'll close this PR and create a new one with the changes accordingly.

Issue #436, opened 2020-06-07
kotlin linting


knyttl:
Hi, when implementing https://github.com/http4k/http4k/pull/435 I noticed the codebase does not have a linter set up - e.g. some files end with new lines and some don't. 

I am not sure whether this is desired, I would be happy to participate if wanted.

nlochschmidt:
I noticed this as well. At least there is an .editorconfig that can be used in IntelliJ to automatically add newlines on save. That said, [ktlint](https://ktlint.github.io/) is what I typically use in my projects and I have made mostly good experiences with it.

daviddenton:
We've used ktlint before on projects. IMO, it's definitely more important on bigger projects where there is a lot of activity, but can get annoying if you've got "all warnings are errors" switched on (see below), which would seem to be the obvious mode.

In truth, we've just never got around to actually putting linting in. I'm happy enough to get it PR'd into the project  as long as it doesn't slow down the build a bunch and we can remove the "error on unused variable" warning - that setting is insanely annoying if you're in the middle of developing and want to run something with one arg unused for any reason. 😢

goodhoko:
I guess this was resolved with https://github.com/http4k/http4k/pull/576. ,)

daviddenton:
Agreed. We can reopen this if we need to! :)

PR #495, opened 2020-10-06
#436 Fix ktlints rules

knyttl:
These are individual commits which fix separate ktlint rules. Only rule I didn't dare fixing is the `filename` rule - this should be handled manually by design.

If you are unhappy with some of the rules, these can be easily disabled and corresponding commit dropped.

The last commit enables ktlint check within the gradle build.

I am not fimiliar with github's pre-push hooks (I am using gitlab), but there should be one implemented as follows:

```
#!/bin/sh
# This is a ktlint pre-commit/pre-push hook to validate the introduced changes to conform to kotlin code style.
# To install this hook, copy this file to .git/hooks with the filename `pre-commit` or `pre-push` as required.
# If you want multiple scripts to run on a single commit hook refer to https://stackoverflow.com/questions/26624368/handle-multiple-pre-commit-hooks
# (originally sourced from: https://github.com/pinterest/ktlint/blob/master/ktlint/src/main/resources/ktlint-git-pre-commit-hook.sh)
DISABLED_RULES='filename'
git diff --name-only HEAD | grep '\.kt[s"]\?$' | xargs ktlint --disabled_rules=$DISABLED_RULES
```


daviddenton:
Thanks v much for the contrib - it's pretty vast so would like to split it down a bit.

Question: How much of this is "linting for style" and how much is "linting for formatting"? For the latter, we definitely don't want to have to rely on manual fomatting for anything - so we really want all of it to be automated via intellij. 

Wondering if we can put ktlint in first with a simple set of rules that we would apply anyway with intellij - eg no * imports + colon/curly spacing, see that it doesn't smash the build performance, and then gradually introduce the new automated rules bit by bit.

knyttl:
You can certainly start with just this:

```
ktlint -F --disabled_rules=no-wildcard-imports,import-ordering,no-consecutive-blank-lines,colon-spacing,curly-spacing,range-spacing,no-var,dump,chain-wrapping,comment-spacing,filename,final-newline,indent,max-line-length,modifier-order,no-blank-line-before-rbrace,no-line-break-after-else,no-line-break-before-assignment,no-semi,no-trailing-spaces,parameter-list-wrapping,colon-spacing,comma-spacing,curly-spacing,op-spacing,no-unused-imports,string-template
```

Which basically disables all the rules. In the beggining you can just enable the ones you mention. However in my experience this does not have any performance implications and you can start with more rules. See the individual commits, some of them make a lot of sense.

knyttl:
Can I be of more help?

daviddenton:
Apologies - this kind of fell off the radar. 

I think if we could get another PR with just the addition of the lintinh added and all rules disabled, then we can look at introducing them one by one.

knyttl:
Closing in favor of https://github.com/http4k/http4k/pull/568

PR #503, opened 2020-10-21
Documentation: permalink added at the end of pages titles

razvn:
Why:
I find usefull to be able to give someone a precise link to a specific part of a page of the documentation.

What:
The modifcation adds a ¶ at the end of pages' titles when the mouse is over it that is a link to that specific title.

How:
I activated the mkdocs toc/permalinks extenesion and added some css.

Result:
image

### DIFF
--- a/src/docs/css/main.css
+++ b/src/docs/css/main.css
@@ -104,6 +104,16 @@ h4,
   font-weight: 700;
 }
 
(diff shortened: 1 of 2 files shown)


s4nchez:
Thank you for the PR, @razvn!

PR #510, opened 2020-10-28
Servers can be described in the contract

### LINKED ISSUE #433: Support OpenApi servers
The [OpenApi specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#openapi-object) allows top-level `servers` object to be specified:

> Field Name | Type | Description
> ---|:---:|---
> <a name="oasServers"></a>servers | [[Server Object](#serverObject)] | An array of Server Objects, which provide connectivity information to a target server. If the `servers` property is not provided, or is an empty array, the default value would be a [Server Object](#serverObject) with a [url](#serverUrl) value of `/`.

As far as I can see we currently cannot control it with http4k.

zsambek:
Fixes #433 
Questions, ideas:
- Do we want to use `Appendable<ServerObject>` instead of `List<ServerObject>`?
- Do we want to add `http://localhost:9000` as the default? 
- Would it be better if we are not exposing the `ServerObject` which is an OpenAPI v3 model?
- Do we want to update the `documenting_apis_with_openapi` tutorial with this new feature? 

What are your thoughts @daviddenton?


zsambek:
It seems parameterized servers are not supported by the OpenAPI Gradle Generator Plugin. I will create a separate unit test just for that.

daviddenton:
Sorry it's taken so long to get to this... 🙄 

>Do we want to use Appendable<ServerObject> instead of List<ServerObject>?
See below...

> Do we want to add http://localhost:9000 as the default?
From the spec: ```An array of Server Objects, which provide connectivity information to a target server. If the servers property is not provided, or is an empty array, the default value would be a Server Object with a url value of /.```

Would it be better if we are not exposing the ServerObject which is an OpenAPI v3 model?
For OpenApi specific things, we can change the OpenApi3 contract renderer constructor to take the Server list. This will avoid us changing the generic contract for specific things.

Do we want to update the documenting_apis_with_openapi tutorial with this new feature?
> absolutely! 😄

goodhoko:
@zsambek Thanks for addressing my issue! Is there anything I can help with to get this merged? I should have some spare time in the coming days. .)

knyttl:
merging this would be very very helpful.

daviddenton:
There are some merges and changes to be made before we can. :)

zsambek:
@daviddenton Could you please take a look at the PR? I just resolved the merge conflicts.

daviddenton:
Just a couple of things:

1. `Do we want to update the documenting_apis_with_openapi tutorial with this new feature?` Yes. :) 
2. Can you sort out the imports? We don't use * imports. You can find our .edtorconfig at the root of the project
3. ServerObject can probably be renamed to `ApiServer` or similar, and the description can default to null. Also let's use a Uri instead of a String for the URL.
:)

daviddenton:
Closing this off as has been implemented in #674

PR #514, opened 2020-11-05
Remove log spam

NersesAM:
This line is logging directly to console on every request. As its not using logging libraries took me a while to track down where it's coming from.


daviddenton:
whoops! Thank you - and apologies for the hassle in tracking it down!

daviddenton:
@NersesAM JTLYK that we've released version 3.273.0 with this fix in.

NersesAM:
Awesome! thanks you @daviddenton

PR #515, opened 2020-11-08
Kotlin serialization auto-marshalling

### LINKED ISSUE #478: Automarshalling with kotlinx serialization
Is this possible with KotlinX and just not implemented? Or is there some reason why it's impossible?

zsambek:
Fixes #478


razvn:
Congrats @zsambek  for the implementation ! I've been playing around a bit with this too and got stucked as couldn't figure out how to make it work with generics... nice catch the `json.serializersModule.serializer(input::class.java)` !

Thanks

zsambek:
@razvn Thanks! :)  
I have to admit it is far from perfect but I think it is a good first step. 
I've wondered if it would be better to avoid the `JsonLibAutoMarshallingJson<JsonElement>()` and implement all the methods on my own so I could use `reified T` for all the methods, but then I decided to do it like this.

Issue #544, opened 2020-12-14
Http4k client with Jetty throws nullpointer exception


Hakky54:
Hi, thank you for this cool library. I like the easy to use api for the http clients. I was playing around and adding examples to my own projects and in that way I discovered that there is a bug within the response mapper of the JettyClient.

The following code within `org.http4k.client.JettyClient` is throwing a nullpointer exception:
```
private fun JettyResponse.toHttp4kResponse(): Response =
    Response(Status(status, reason)).headers(headers.toHttp4kHeaders())
```

This happens when the Server is not returning a reason phrase.

See here for a screenshot:
<img width="1057" alt="Screenshot 2020-12-14 at 00 58 08" src="https://user-images.githubusercontent.com/16032204/102027910-7c596c00-3da7-11eb-8395-a5ddb07d8dd5.png">

I discovered that the some servers don't always send the reason phrase. I am using Tomcat and they mentioned it is disabled by default, but can be enabled. However this option is deprecated and the switch will be removed for the next release.

I found the following rfc http specification for the reason phrase, https://tools.ietf.org/html/rfc7230#section-3.1.2

>The reason-phrase element exists for the sole purpose of providing a
textual description associated with the numeric status code, mostly
out of deference to earlier Internet application protocols that were
more frequently used with interactive text clients. A client SHOULD
ignore the reason-phrase content.

A possible solution would be maybe using a look up table/map when the reason phrase is null. In that way the status code can be used to grab a reason phrase from the `org.http4k.core.Status class` for example. If there is no matching status code within the lookup table we can maybe provide a `unknown reason` or `empty reason`?

s4nchez:
Good spot, @Hakky54 . Thank you for reporting this.

There are many scenarios where either clients or servers don't use status description, but we clearly missed that one.

For now, I believe the simplest fix is to use ".orEmpty" for the reason in that particular client (PR welcome). That'd match the handling done in other places.

In the long term, we should consider either making the status description nullable or have a specific tiny type that can also represent the "no description" value. That'd be a breaking change so we need to have a think/play first.

Hakky54:
Yes it is for fixable. I did an attempt here: https://github.com/http4k/http4k/pull/545

It uses the description if present, or else it will try to find a matching description for the given status code or fall back to a default. What do you of the solution?

daviddenton:
@Hakky54  3.283.1 is being released now with this bug fixed.

Hakky54:
[...]

PR #545, opened 2020-12-14
Bugfix: add nullsafety and fallback for description within the Status

Hakky54:
This PR fixes the following issue: https://github.com/http4k/http4k/issues/544


s4nchez:
Thank you for the PR, @Hakky54 .

The null safety part seems ok. However, could we skip the fallback logic for now, please?

One of the http4k principles is trying to map request and response as "raw" as possible. This make troubleshooting integrations much simpler and forces behaviour to be always defined in filters, which are optional to library clients.

Hakky54:
Sure, it makes the PR much simpler. I just removed it

s4nchez:
Excellent, @Hakky54. Thank you again for the bug report and the PR.

We'll merge this and include in the next release.

Hakky54:
Thank you very much for replying this quickly and approving the PR!

PR #547, opened 2020-12-18
Start ApacheAsyncClient by default if not started yet

Hakky54:
I noticed that the `ApacheAsyncClient` requires a different kind of initialisation compared to the `Apache4AsyncClient`. Before supplying a client to the constructor of the ApacheAsyncClient it needs to be started with the function `client.start()`. This additional call is however not needed when using the `Apache4AsyncClient`.

Although it not a big deal to call the additional start method, but it just caught my attention and that the configuration is not similar. I analysed the source code and discovered that the `Apache4AsyncClient` calls internally `if (!client.isRunning) client.start()` on line number 34. This part is missing within the `ApacheAsyncClient`, but it is easy to include.


Hakky54:
This PR should resolve the following issue: https://github.com/http4k/http4k/issues/548

daviddenton:
Good spot! Thanks @Hakky54

PR #552, opened 2021-01-05
AccessTokenFetcher initializes all AccessToken fields.

paraseba:
We were ignoring fields like the refresh token.


daviddenton:
Good catch. Thanks!

PR #556, opened 2021-01-11
See_Other (303) is a special case of redirection.

dgliosca:
A special case of redirection should be taken into account for See_Other (303).
See_Other should be followed by a GET and as per RFC a GET with body has no defined semantics. See below:

https://tools.ietf.org/html/rfc7231#page-24
_A payload within a GET request message has no defined semantics;
sending a payload body on a GET request might cause some existing
implementations to reject the request._

This PR intends to strips out payload coming from a POST requests that is followed by a GET


s4nchez:
Thank you for the PR, @dgliosca. I'll merge it and include it in the next release 🙂

PR #557, opened 2021-01-13
docs: serializing for a Response

2x2xplz:
added an example of using a BiDi lens for serializing a data class and creating a Response object


daviddenton:
Thanks! :)

PR #560, opened 2021-01-16
Add a new `parent_id` even if a previous one wasn't previously set

s4nchez:


razvn:
You did it faster 😉.

Next time I check if it's something as easy to fix I won't bother you with it.

daviddenton:
Are we sure this is the right thing to do?

Worrying thabwere creating a span which doesn't actually exist...

s4nchez:
The span needs to be created at some point, and is still created in the same place. The only thing that changed was the assumption that it should only be created if a previous parent existed, which I believe is wrong.That meant our client would never start propagating spans unless something external set the initial parent. Was that the expected behaviour?

I've added a [new test](https://github.com/http4k/http4k/blob/24d87f0f6ff10f080b625a9b16678cea0e2526e6/http4k-core/src/test/kotlin/org/http4k/filter/RequestTracingTest.kt#L50) to demonstrate what I expect.

PR #564, opened 2021-01-22
Generate javax.validation schema for annotated data classes

knyttl:
This is a work-in-progress PR and I would very much like some feedback on how to handle this.

Please review by commits which I tried to make descriptive as much as possible.

Basically I would like to create a new http4k module which would provide the `javax-validation` processing. However I didn't find an easy way how to plug the functionality into `AutoJsonToJsonSchema.kt`, so I duplicated it as `AutoJsonToJsonValidatedSchema.kt` and did the required changes.

Questions:
1. If this is polished according to your standards, is it acceptable for you to add as a new module like I am proposing?
1. How to implement the `AutoJsonToJsonValidatedSchema.kt` without copy-pasting the original `AutoJsonToJsonSchema.kt`?
1. I guess it is not okay to directly modify the the `FieldMetadata` class with `javax-validation` related fields. How should this be handled? We  could make it an interface which I could implement, but the problem would remain as its payload needs to be passed to `SchemaNode.Primitive` :/
1. I am not sure how to refer to correct kotlin versions from the new module.

The resulting doc looks like this – it would make sense to instruct jackson to ignore the `null` fields in this case.
Screenshot 2021-01-22 at 14 10 42

CC @goodhoko

### DIFF
--- a/http4k-security-oauth/src/main/kotlin/org/http4k/security/oauth/testing/FakeOAuthServer.kt
+++ b/http4k-security-oauth/src/main/kotlin/org/http4k/security/oauth/testing/FakeOAuthServer.kt
@@ -40,14 +40,14 @@ object FakeOAuthServer {
         authPath: String,
         tokenPath: String,
         clock: Clock = Clock.systemDefaultZone(),
-        name: String = "OAUTH"
+        accessCodePrefix: (AuthorizationCode) -> String = { "OAUTH_" + it.value.reversed() }
     ): HttpHandler {
         val server = OAuthServer(
             tokenPath,
             InMemoryAuthRequestTracking(),
             AlwaysOkClientValidator(),
             InMemoryAuthorizationCodes(clock),
-            SimpleAccessTokens(name),
+            SimpleAccessTokens(accessCodePrefix),
             Jackson,
             clock
         )
@@ -110,7 +110,7 @@ private class InMemoryAuthRequestTracking : AuthRequestTracking {
         }
 }
 
-private class SimpleAccessTokens(private val prefix: String) : AccessTokens {
+private class SimpleAccessTokens(private val prefix: (AuthorizationCode) -> String) : AccessTokens {
     override fun create(clientId: ClientId, tokenRequest: TokenRequest) =
         Failure(UnsupportedGrantType("client_credentials"))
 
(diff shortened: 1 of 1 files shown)


s4nchez:
Hi, @knyttl . I like what this new module could enable. However, I also see the challenge of plugging it straight into the JSON integration.

Wouldn't it be possible to build the validation _on top_ of the existing `AutoJsonToJsonSchema` (i.e. deserialised into an intermediate/invalidated representation before applying the validation)? That way you could potentially also decouple the validation from JSON and apply something similar to anything lens-based (query/path/form etc)

knyttl:
> Wouldn't it be possible to build the validation _on top_ of the existing `AutoJsonToJsonSchema` (i.e. deserialised into an intermediate/invalidated representation before applying the validation)? That way you could potentially also decouple the validation from JSON and apply something similar to anything lens-based (query/path/form etc)

Could you extend/elaborate a bit more your idea? I am not sure how this would be done. 

The problem is that the original class uses a mostly extensions and it cannot be directly/easily overriden - and mostly these are private (which can be changed though). The problem lies mostly with the `FieldMetadata` and `Primitive` classes, which would need to contain the new fields :/

s4nchez:
What I mean is decoupling validation from deserialisation. Let the JSON->Object work as it is and then add a layer of Object->ValidatedObject. I'm talking at a high level though, as we'd need to have a play to see the ramifications of that approach.

daviddenton:
There is the option of using the extensions API, which gives you direct access to the JSON structure.

:)

/dave

knyttl:
>  @s4nchez Let the JSON->Object work as it is and then add a layer of Object->ValidatedObject

The problem is that the `JSON->Object` looses the information about validation - it is available only on the kclass level. Just to confirm we are talking about this part:

```
private fun NODE.toObjectSchema(objName: String?, obj: Any, isNullable: Boolean, topLevel: Boolean, metadata: FieldMetadata?): SchemaNode.Reference {
        val properties = json.fields(this)
```

I guess we would have to replace mapping of the json (created from the data class) with direct mapping of the data class?

> @daviddenton extensions API, which gives you direct access to the JSON structure.

The problem is that these annotations are not available on the JSON level, notice I need to access it via reflection:

```kotlin
val dataClassField: KCallable<*>? = obj::class.members.find { it.name == fieldName }
val size = dataClassField.findGetterAnnotation<Size>()
```

A little bit off topic - would you maybe consider taking out the OpenApi generator out of `http4k-context` to `http4k-openapi`? that might allow better forking of this part.

daviddenton:
I suggest that the only field added to the FieldMetadata is a Map<String, String>, representing all of the extra fields which should be merged into the field definition. This will minimise the changes required to the schema generator. The map can be populated by a custom FieldRetreival which can be inserted into the chain when constructing the OpenApi instance

knyttl:
I wanted to push from a different branch, but github just closed the pull requested as I force-pushed my master. I will try as @daviddenton proposes and create a new PR.

Issue #541, opened 2020-12-10
javax.validation API


knyttl:
Spring MVC supports javax.validation annotations `@Size`, `@Pattern` which allows:

1. Validate input fields on data classes
2. Generate OpenApi/swagger docs, eg.:
   <img width="711" alt="Screenshot 2020-12-10 at 14 52 55" src="https://user-images.githubusercontent.com/401107/101780891-6cb3fc00-3af7-11eb-9807-65966cd1e03d.png">

These can be used on data classes as follows:

```kotlin
   data class User(
        @get:Size(min = 0, max = 64)
        val name: String?,
        @get:Size(min = 0, max = 512)
        val bio: String?,
    )
```

It would be great if http4k could get use of these annotations for both reasons above.

PR #567, opened 2021-01-30
#541 Generate javax.validation schema for annotated data classes^2

knyttl:
This is a follow up MR for https://github.com/http4k/http4k/pull/564, please review by individual commits.

Based on @daviddenton's suggestion for FieldMetadataRetrieval, I refactored the previous code. Probably the only remaining issue is flattening the `SchemaNode.extra` so that its contents become directly part of the `SchemaNode`.

Now the resulting node looks like:

```json
"bio": {
   "example": "Bar",
   "description": null,
   "extra": {
        "minLength": 0,
        "maxLength": 512
    },
    "type": "string"
},
```

but to be compliant, it should look like:

```json
"bio": {
   "example": "Bar",
   "description": null,
   "minLength": 0,
   "maxLength": 512,
   "type": "string"
},
```

CC @goodhoko

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
@@ -68,7 +68,7 @@ object JacksonFieldMetadataRetrievalStrategy : FieldMetadataRetrievalStrategy {
     override fun invoke(target: Any, fieldName: String): FieldMetadata =
         FieldMetadata(description = target.javaClass.findPropertyDescription(fieldName))
 
-    private fun Class<Any>.findPropertyDescription(name: String): String? =
+    fun Class<Any>.findPropertyDescription(name: String): String? =
         kotlin.constructors.first().parameters
             .firstOrNull { p -> p.kind == KParameter.Kind.VALUE && p.name == name }
             ?.let { p ->
--- a/http4k-javax-validation/build.gradle
+++ b/http4k-javax-validation/build.gradle
@@ -1,0 +1,13 @@
+description = 'Http4k javax-validation support'
+
+dependencies {
+    api project(":http4k-core")
+    api project(":http4k-server-jetty")
+    api project(":http4k-contract")
+    api project(":http4k-format-jackson")
+
+    implementation "javax.validation:validation-api:2.0.1.Final"
+    implementation "io.swagger.core.v3:swagger-annotations:2.1.7"
+    testImplementation project(path: ":http4k-core", configuration: 'testArtifacts')
+    testImplementation Config.TestDependencies
+}
--- a/http4k-javax-validation/build.gradle
+++ b/http4k-javax-validation/build.gradle
@@ -2,9 +2,11 @@ description = 'Http4k javax-validation support'
 
 dependencies {
     api project(":http4k-core")
-    api project(":http4k-server-jetty")
     api project(":http4k-contract")
     api project(":http4k-format-jackson")
+
+    // TODO remove with demo
+    api project(":http4k-server-jetty")
 
     implementation "javax.validation:validation-api:2.0.1.Final"
     implementation "io.swagger.core.v3:swagger-annotations:2.1.7"
--- a/http4k-javax-validation/src/test/kotlin/org/http4k/javaxvalidation/JavaXValidationFieldRetrievalStrategyTest.kt
+++ b/http4k-javax-validation/src/test/kotlin/org/http4k/javaxvalidation/JavaXValidationFieldRetrievalStrategyTest.kt
@@ -1,0 +1,49 @@
+package org.http4k.javaxvalidation
+
+import com.fasterxml.jackson.annotation.JsonPropertyDescription
+import io.swagger.v3.oas.annotations.media.Schema
+import org.http4k.contract.openapi.v3.FieldMetadata
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import javax.validation.constraints.Max
+import javax.validation.constraints.Min
+import javax.validation.constraints.Pattern
+import javax.validation.constraints.Size
+
+internal class JavaXValidationFieldRetrievalStrategyTest {
+
+    @Test
+    fun testJacksonDescription() {
+        data class NameWithJacksonDescription(@JsonPropertyDescription("Just a description") val name: String)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(NameWithJacksonDescription("Jane Doe"), NameWithJacksonDescription::name.name)
+        assertEquals(FieldMetadata("Just a description", emptyMap()), result)
+    }
+
+    @Test
+    fun testSwagger3Description() {
+        data class NameWithSwagger3Description(@Schema(description = "Just a description") val name: String)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(NameWithSwagger3Description("Jane Doe"), NameWithSwagger3Description::name.name)
+        assertEquals(FieldMetadata("Just a description", emptyMap()), result)
+    }
+
+    @Test
+    fun testLength() {
+        data class NameWithSize(@get:Size(min = 2, max = 64) val name: String)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(NameWithSize("Jane Doe"), NameWithSize::name.name)
+        assertEquals(mapOf("minLength" to 2, "maxLength" to 64), result.extra)
+    }
+
+    @Test
+    fun testMinMaxNumber() {
+        data class IntBounded(@get:Min(30) @get:Max(50) val count: Int)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(IntBounded(40), IntBounded::count.name)
+        assertEquals(mapOf("min" to 30L, "max" to 50L), result.extra)
+    }
+
+    @Test
+    fun testPattern() {
+        data class StringWithPattern(@get:Pattern(regexp = "[a-z]+@[a-z+.]") val email: String)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(StringWithPattern("jane@doe.com"), StringWithPattern::email.name)
+        assertEquals(mapOf("pattern" to "[a-z]+@[a-z+.]"), result.extra)
+    }
+}
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/AutoJsonToJsonSchema.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/AutoJsonToJsonSchema.kt
@@ -1,5 +1,6 @@
 package org.http4k.contract.openapi.v3
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter
 import org.http4k.format.AutoMarshallingJson
 import org.http4k.format.JsonType
 import org.http4k.lens.ParamMeta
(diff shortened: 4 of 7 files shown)


http4k-javax-validation/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+dependencies {
+    api project(":http4k-core")
+    api project(":http4k-server-jetty")
+    api project(":http4k-contract")
+    api project(":http4k-format-jackson")
+
+    implementation "javax.validation:validation-api:2.0.1.Final"
+
+    // TODO remove everything below
```
daviddenton, line 11: Can we remove this now?

http4k-javax-validation/src/main/kotlin/org/http4k/javaxvalidation/OpenApiTest.kt
```diff
@@ -0,0 +1,82 @@
+package org.http4k.javaxvalidation
```
daviddenton, line 1: This isn't a test.  :)

daviddenton:
There isn't test coverage for any of the code. I'm thinking of this as 2 changes- the changes to the contract api to support the validation change, and then  work out what to do with the validation retreival stuff itself..

http4k-javax-validation/src/main/kotlin/org/http4k/javaxvalidation/OpenApiTest.kt
knyttl: Renamed to demo.

http4k-javax-validation/build.gradle
knyttl: removed

knyttl:
> There isn't test coverage for any of the code, but the approach is fine I think. 

Tests added.

> I'm thinking of this as 2 changes- the changes to the contract api to support the validation change, and then work out what to do with the validation retreival stuff itself..

For this reason, it is split in two separate commits.

> But overall if we can get the metadata into the right shape for merging into the field structure then people can write their own metadata plugins to support whatever they need.

Here I am not sure what would be a correct approach.

daviddenton:
At the moment, the problem of merging extra fields into the overall structure is unsolved, which is a major problem. Line 66 of the AutoJsonToJsonSchema needs to be able to construct a SchemaNode with not just the properties of the sealed class instance, but also the map of extra properties.

knyttl:
> At the moment, the problem of merging extra fields into the overall structure is unsolved, which is a major problem. Line 66 of the AutoJsonToJsonSchema needs to be able to construct a SchemaNode with not just the properties of the sealed class instance, but also the map of extra properties.

I think we can live with the state as here - the result is still nicely visible and also parsable (we process the generated JSON for frontend validation).

Maybe we can leave it like this.

daviddenton:
afraid we can't leave it like this because the output of the renderer will not be accessible as "extra" (which doesn't actually do anything anyway btw so can be removed), and if we make Schema.metadata a val it will appear on every node.

http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
```diff
@@ -66,12 +66,12 @@ class JacksonJsonNamingAnnotated(private val json: ConfigurableJackson = Jackson
 
 object JacksonFieldMetadataRetrievalStrategy : FieldMetadataRetrievalStrategy {
     override fun invoke(target: Any, fieldName: String): FieldMetadata =
-        FieldMetadata(description = target.javaClass.findPropertyDescription(fieldName))
+        FieldMetadata(description = findPropertyDescription(target.javaClass, fieldName))
 
-    private fun Class<Any>.findPropertyDescription(name: String): String? =
-        kotlin.constructors.first().parameters
+    fun findPropertyDescription(clazz: Class<Any>, name: String): String? =
```
goodhoko, line 71: AFAIK there's no need to change this from extension to 'regular' function. Just marking this as public (un-marking as private) should be enough to make this callable from other places.

Personally I too prefer the regular format over extension, but here I'd advocate for smaller diff.

knyttl:
> afraid we can't leave it like this because the output of the renderer will not be accessible as "extra" (which doesn't actually do anything anyway btw so can be removed), and if we make Schema.metadata a val it will appear on every node.

Okay, is there something I can do or will you have a look at it? (in some future)

http4k-javax-validation/src/test/kotlin/org/http4k/javaxvalidation/JavaXValidationFieldRetrievalStrategyTest.kt
```diff
@@ -0,0 +1,32 @@
...
+        assertEquals(mapOf("min" to 30L, "max" to 50L), result.extra)
+    }
+
+    @Test
+    fun testPattern() {
+        data class StringWithPattern(@get:Pattern(regexp = "[a-z]+@[a-z+.]") val email: String)
+        val result = JavaXValidationFieldRetrievalStrategy.invoke(StringWithPattern("jane@doe.com"), StringWithPattern::email.name)
+        assertEquals(mapOf("pattern" to "[a-z]+@[a-z+.]"), result.extra)
+    }
+}
```
goodhoko, line 32: It would be cool to also test this integrated into a contract.

goodhoko (commented):
I don't have the proposition to act as a reviewer here but I'll throw in my two cents. .)

I really like what this feature aims to enable. What I'd expect from this module is that just by decorating input/output data classes with javax.validation annotations we'll get
1. Automatic assertion of the rules when extracting the data from request/response and serving BAD_REQUEST if the rules are not met.
2. Automatic assertion when injecting the data into request/response. Although, maybe this should happen when constructing the data classes? I'm not sure what are the possibilities with javax.validation here.
3. Automatic generation of appropriate OpenApi documentation for the annotated fields.

I think that the minimal viable feature set is [1, 3] - that's enough for the the typical use case of validating API input. However, @daviddenton might have different opinion on what's a typical use case. .)

As far as I understand this MR only implements 3., right? What are the plans or possible ways to implement 1. and 2.?

knyttl:
> I don't have the proposition to act as a reviewer here but I'll throw in my two cents. .)
> 
> I really like what this feature aims to enable. What I'd expect from this module is that just by decorating input/output data classes with javax.validation annotations we'll get
> 
> 1. Automatic assertion of the rules when extracting the data from request/response and serving BAD_REQUEST if the rules are not met.
> 2. Automatic assertion when injecting the data into request/response. Although, maybe this should happen when constructing the data classes? I'm not sure what are the possibilities with javax.validation here.
> 3. Automatic generation of appropriate OpenApi documentation for the annotated fields.
> 
> I think that the minimal viable feature set is [1, 3] - that's enough for the the typical use case of validating API input. However, @daviddenton might have different opinion on what's a typical use case. .)
> 
> As far as I understand this MR only implements 3., right? What are the plans or possible ways to implement 1. and 2.?

if this gets merged, I will propose the 1 & 2 as well. I didn't want to spent time implementing something that might not be accepted in the end. I can see @daviddenton is currently pushing some related commits, so I guess we will eventually get there.

http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt
knyttl: Unfortunately it is, otherwise it is not usable from the `JavaXValidationFieldRetrievalStrategy`. See it is imported and used there.

goodhoko: I do see that. I wasn't sure before writing this comment so I pulled your code in a tried myself. You can happily import and use extension functions from other packages as long as they're public.

knyttl: Interesting, for some reason it didn't work for me when implementing it. Will fix.

daviddenton:
It's worth pointing out here that rule 5 of the http4k [ethos[(https://www.http4k.org/rationale/#about) is:
> **Absolutely no magic involved: No reflection. No annotations.**

We have already bent this rule with the Jackson annotations (to deal with the shortcomings of Jackson TBH), but relying on annotations as a part of the library functionality is a no-no. from us. We will, however do what we can to let people choose to extend http4k in their own arenas with simple API changes.

As well as being against our own rules, incorporating javax.validation annotations into the Lens system would require a total rewrite and would introduce an external dependency in the core, so absolutely will not happen. We recommend enforcing validation rules in expresable code and not with magic which is why the OpenApi generation is designed as it is. Hence, the total javax.valdation roundtripping is not a viable goal.

The changes I made to the AutoJsonToJsonSchema were to dry it out a bit and I've confirmed that this is not a simple change to merge in the metadata that I can see. The challenge is to get the SchemaNode FieldMetadata properties merged in with the other fields, which requires a change in the SchemaNode object hierarchy, and in a way that does not bind the API to Jackson. 

Ideas welcome on how to achieve that.. 😄

knyttl:
> This one is a bit tricky. I need to be able to select the tickets I want to refund - but I don't want to complicate the refund modal, as that might have some ugly edge cases.

I guess this is a great and very good statement for the core http4k. I guess optional secondary modules might break this rule if noone is forced to use them.

> Ideas welcome on how to achieve that.. 😄

My only idea (naive) idea is something like:

```kotlin
openapiPayload = jackson.convert(schemaNode, Map::class) + extra
```

knyttl:
> > Absolutely no magic involved: No reflection. No annotations.
> 
> I guess this is a great and very good statement for the core http4k. I guess optional secondary modules might break this rule if noone is forced to use them.
> 
> > Ideas welcome on how to achieve that.. 😄
> 
> My only idea (naive) idea is something like:
> 
> ```kotlin
> openapiPayload = jackson.convert(schemaNode, Map::class) + extra
> ```
> 
> Edit:
> 
> I know this is again another annotation, but this could also be resolved with `@JsonAnyGetter` annotation:
> 
> ```
> private sealed class SchemaNode {
>    ...
>    @JsonAnyGetteer
>    val extra = metadata?.extra
> ```
> 
> which would make the extras to be serializable on the same level as SchemaNode.

@daviddenton what do you think about these two options?

knyttl:
I experimented with it again and the `@get:JsonAnyGetter` is probably the only solution - it flattens the extra to the output map correctly:

```kotlin
 private sealed class SchemaNode {
   ...
   @get:JsonAnyGetteer
   val extra = metadata?.extra
```
and the result:

```json 
"name": {
  "example": "Foo",
  "description": "Name of the entity.",
  "type": "string",
  "minLength": 0,
  "maxLength": 64
},
```

so - trying one last time 🙂: even if you are using `@JsonPropertyDescription` from Jackson, you still want to avoid using `@JsonAnyGetter` from the same package? - it is already included and no new dependency is needed. This would solve everything 🙏

daviddenton:
This functionality is now at least partly possible due to the fact we now support the various other types of OpenAPI fields in SchemaNode. All you need to do is to add a `FieldMetadataRetrievalStrategy` to populate the fields. The SchemaNode now contains this:

```kotlin
    val description = metadata?.extra?.get("description")
    val format = metadata?.extra?.get("format")
    val default = metadata?.extra?.get("default")
    val title = metadata?.extra?.get("title")
    val multipleOf = metadata?.extra?.get("multipleOf")
    val maximum = metadata?.extra?.get("maximum")
    val exclusiveMaximum = metadata?.extra?.get("exclusiveMaximum")
    val minimum = metadata?.extra?.get("minimum")
    val exclusiveMinimum = metadata?.extra?.get("exclusiveMinimum")
    val maxLength = metadata?.extra?.get("maxLength")
    val minLength = metadata?.extra?.get("minLength")
    val pattern = metadata?.extra?.get("pattern")
    val maxItems = metadata?.extra?.get("maxItems")
    val minItems = metadata?.extra?.get("minItems")
    val uniqueItems = metadata?.extra?.get("uniqueItems")
    val maxProperties = metadata?.extra?.get("maxProperties")
    val minProperties = metadata?.extra?.get("minProperties")
```

Am going to close this PR now, but thought it would be useful to let you know where we are @knyttl  😄

PR #568, opened 2021-01-31
#436 Fix ktlints rules

knyttl:
This is a follow up for https://github.com/http4k/http4k/pull/495.

I enabled just one rule (final-newline) for demonstration.


knyttl:
> Ivan Sanchez [http4k]  8:23 PM
> I looked at the ktlint one earlier today. What did you use to fix the issues automatically? Is that documented in the PR?

Added to the comment in `.editorconfig` (use ktlint -F)

> Also, is there a report or something we could publish regarding linting? I’m thinking we should add that as a separate (non breaking) build step to start with.

Not sure what you mean.

> Also, does it include ide config so formatting in IntelliJ is enough to prevent linting issues?

Added in .editorconfig

Ideally it would be great to go through all the rules and decide which should be enabled. There are some, which might be interesting:

- import-ordering
- no-consecutive-blank-lines
- colon-spacing
- comma-spacing
- no-unused-imports

knyttl:
ping :-)

s4nchez:
> Added to the comment in `.editorconfig` (use ktlint -F)

👍 

> 
> > Also, is there a report or something we could publish regarding linting? I’m thinking we should add that as a separate (non breaking) build step to start with.
> 
> Not sure what you mean. Basically it should be run as precommit hook (see tools directory) and it is run as a gradle task during build.

We're not interested in enforcing a precommit hook. Our goal is to keep the developer experience as simple as possible, and that means anyone can just checkout and is ready to build locally and push.

> Added in .editorconfig

👍 

> Ideally it would be great to go through all the rules and decide which should be enabled. There are some, which might be interesting:
> 
> * import-ordering
> * no-consecutive-blank-lines
> * colon-spacing
> * comma-spacing
> * no-unused-imports

At this stage, I'm inclined to cherry-pick only your fist commit (add ktlint with everything disabled) and see when/how we can best run it before enabling any rule.

Finally, sorry for this taking a while. We've been busy with other work/life commitments lately. And thank you for your contributions. It's really appreciated.

knyttl:
> We're not interested in enforcing a precommit hook. Our goal is to keep the developer experience as simple as possible, and that means anyone can just checkout and is ready to build locally and push.

Ok, the pre-commit hook cannot be enforced, so it is up to everyone if they set it up.

However. If you still want to allow commits which violate the allowed rules, the gradle plugin should be disabled. But I think that would be a pity, if you allow committing code which doesn't follow the rules you set up.

s4nchez:
Closing as linting has been added as part of #576

PR #571, opened 2021-02-03
Add a basic etag filter

jshiell:
I couldn't find an eTag handler in the default code. So this is a port of what we're using internally.

* Takes the etag from the response
* If absent, generates a strong etag by hashing the response when we have the response in memory (i.e. it's not streaming)
* Checks this against `if-none-match`, if present, and returns a 304 if appropriate
* Otherwise attaches the etag to the response

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
@@ -4,9 +4,13 @@ import org.http4k.core.ContentType
 import org.http4k.core.Filter
 import org.http4k.core.HttpHandler
 import org.http4k.core.HttpTransaction
+import org.http4k.core.Method
 import org.http4k.core.Request
 import org.http4k.core.Response
+import org.http4k.core.Status
 import org.http4k.filter.GzipCompressionMode.Memory
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
 import java.time.Clock
 import java.time.Duration
 import java.time.Duration.between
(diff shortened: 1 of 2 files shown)


http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
```diff
@@ -131,6 +135,74 @@ object ResponseFilters {
     fun Base64EncodeBody() = Filter { next ->
         { next(it).run { body(Base64.getEncoder().encodeToString(body.payload.array())) } }
     }
+
+    class EtagSupport : Filter {
+        private val md5 = ThreadLocal.withInitial {
+            try {
+                MessageDigest.getInstance("MD5")
+            } catch (e: NoSuchAlgorithmException) {
```
knyttl, line 143: Why would this need catching? Would it make any difference as this is not Java anymore?

http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
```diff
@@ -131,6 +135,74 @@ object ResponseFilters {
...
+
+        private fun Response.headerIfPresent(name: String, value: String?) = if (value != null) {
+            this.header(name, value)
+        } else {
+            this
+        }
+
+        private fun generateStrongEtag(content: ByteArray): String = "\"${hash(content).toHex()}\""
+
+        private fun hash(bytes: ByteArray): ByteArray = md5.get().digest(bytes)
```
knyttl, line 197: why here you are using regular function and below extension? maybe this could be unified?

http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
jshiell: It's a leftover from before the optimisation to make it a supplier to a threadlocal went in. Feel free to drop it.

http4k-core/src/main/kotlin/org/http4k/filter/ResponseFilters.kt
jshiell: As said, this is a port from our internal code, which was original ported from Java and Utterlyidle, and has been modified by multiple people over the years. Upside, it's prod tested. Downside, there are inconsistencies. Free free to clean it up, it could probably do with it.

jshiell: Done: https://github.com/http4k/http4k/commit/0365ad2a11a443f4a665c2dc871b62e98dedf7f0

PR #576, opened 2021-02-10
Add ktlint plugin with most of the rules disabled

s4nchez:
Cherry-picked from #568


knyttl:
I think the editorconfig is missing the `final-newline` rule disabled

s4nchez:
@knyttl I couldn't disable rules via `.editorconfig`. I had to disable them directly in `build.gradle`.

Any ideas? I now have the same rules listed in both files, but if you remove them from `build.gradle`, ktlint fails...

knyttl:
> @knyttl I couldn't disable rules via `.editorconfig`. I had to disable them directly in `build.gradle`.
> 
> Any ideas? I now have the same rules listed in both files, but if you remove them from `build.gradle`, ktlint fails...

It is caused by this issue: https://github.com/pinterest/ktlint/issues/739 and in fact by https://github.com/editorconfig/editorconfig/issues/430. Basically when you add `final-newline`, the line passes character limit for value and will be silently ignored.

I guess the quick&dirty solution would be to decide for some rules and enable them. I don't think it should be such an issue if PRs like https://github.com/http4k/http4k/pull/574 are accepted.

goodhoko:
> It is caused by this issue: pinterest/ktlint#739

The fix for that issue is already merged and waiting for release. I think we can just wait for ktlint to relase a new version. I tried to push that forward [here](https://github.com/pinterest/ktlint/pull/1042#issuecomment-777382964). .)

s4nchez:
@knyttl update on this:

* Added ktlint to all modules (it was previously checking only the root module). 
* Enabled and fixed just the most basic, space-related rules.
* Using just .editorconfig to configure disabled rules as it's now shorter than 255 chars and won't hit the Ktlint bug.

Unless you can spot anything I'm missing, I'm happy to merge this (and close the other PR) and see the impact on master before enabling more rules.

.editorconfig
```diff
@@ -18,6 +18,5 @@ ij_continuation_indent_size = 4
 ij_continuation_indent_size = 4
 ij_kotlin_name_count_to_use_star_import = 99
 ij_kotlin_name_count_to_use_star_import_for_members = 99
-# These are currently disabled ktlint rules, used for: git precommit hook (see `tools/ktlint-precommit-hook.sh`),
-# gradle ktlint check (see `build.gradle`) and manual reformatting with `ktlint -F`.
-disabled_rules = indent,no-wildcard-imports,import-ordering,no-consecutive-blank-lines,colon-spacing,curly-spacing,range-spacing,no-blank-line-before-rbrace,chain-wrapping,comment-spacing,parameter-list-wrapping,comma-spacing,no-unused-imports,string-template,filename
+disabled_rules=chain-wrapping,colon-spacing,comma-spacing,comment-spacing,curly-spacing,filename,final-newline,import-ordering,indent,indent_size,no-blank-line-before-rbrace,no-consecutive-blank-lines,no-multi-spaces,no-trailing-spaces,no-unused-imports,no-wildcard-imports,parameter-list-wrapping,range-spacing,string-template
```
knyttl: removing the spaces around `=` makes it inconsistent with rest of the file

knyttl:
brilliant! just one nit pick

knyttl (approved):

.editorconfig
```diff
@@ -18,6 +18,5 @@ ij_continuation_indent_size = 4
 ij_continuation_indent_size = 4
 ij_kotlin_name_count_to_use_star_import = 99
 ij_kotlin_name_count_to_use_star_import_for_members = 99
-# These are currently disabled ktlint rules, used for: git precommit hook (see `tools/ktlint-precommit-hook.sh`),
```
knyttl: I agree the comment here is probably not a good place, but maybe this could be moved to some more relevant location, maybe some dev guidelines?

s4nchez: I'm not sure about this one. Like I mentioned on the other PR, we don't want to ask people to do more than checkout to be able to work on our project. Also, ktlintCheck runs as part of the normal Gradle build, and people should build locally before pushing. If they don't, that'll fail in PR builds anyway.

So, for now, I'll keep the hook and the comment as hints for those who want, and we'll see how that works on master. I'm happy to change it later if linting becomes more of a pain.

PR #579, opened 2021-02-15
Fix disappearing description in BiDiLensSpec.multi.defaulted

knyttl:
The proposal is hopefully self-explanatory :-)


daviddenton:
Thanks!

Issue #580, opened 2021-02-16
Directory resource loader can serve files from outside root


daviddenton:
As proven by this test: 

```kotlin
    @Test
    fun `directory loader should not load resources above the base directory`() {
        assertThat(Directory("./src/test/resources").load("../../../../.java-version"), absent())
    }
```

daviddenton:
Fixed and released in v4.3.2.1 (which is quite a pleasing version number! 😄 )

PR #582, opened 2021-02-16
Make resource path check a bit more robust

s4nchez:


s4nchez:
This extends the previous fix to #580

PR #583, opened 2021-02-16
Added interfaces to allow extending RoutedRequest and RoutedResponse via delegates

jenarros:
This PR adds interfaces `RequestWithRoute` and `ResponseWithRoute` to `RoutedRequest` and `RoutedResponse`. This allows extending the capabilities of the `RoutedRequest / RoutedResponse` via delegates such as:

```
class ExtendedRequest(val delegate: RoutedRequest) : RequestWithRoute by delegate {
  ...
}
```

Without these interfaces, extensions such as Request.path would fail to identify that such request is a wrapper around a RoutedRequest that supports path extraction.

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/HttpTransaction.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/HttpTransaction.kt
@@ -1,6 +1,7 @@
 package org.http4k.core
 
-import org.http4k.routing.RoutedRequest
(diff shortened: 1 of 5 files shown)


jshiell:
I heartily endorse this product and/or service.

jshiell:
Thanks @s4nchez !

jenarros:
Thanks!

PR #592, opened 2021-03-04
fix(http4k-aws): add x-amz-content-sha256 to SignedHeaders

tkint:
Hello!

This PR is a quick fix for custom domain usage of the module `http4k-connect` which uses `http4k-aws`: https://github.com/http4k/http4k-connect/issues/17
Unfortunately, I cannot try with a real `amazonaws.com` s3 service since I don't have any at my disposal, but I tried this fix on our on-premise internal installation and it works.


s4nchez:
Thank you for the PR, @tkint. 

I've checked it against s3.amazonaws.com using the `AwsRealTest` and it works fine.

PR #613, opened 2021-05-03
FlashAttributesFilter.kt is a Filter that allow Flash Attributes

uberto:
- so that is possible to pass a String after a redirect but only once. It's based on a Request Cookie.


s4nchez:
Apart from the ktlint violations (you can use `./gradlew ktlintCheck` and `./gradlew ktlintFormat` to work around those), it looks ok to me.

PR #620, opened 2021-05-18
Multi-frame websocket messages are handled in Jetty

endofhome:
This fixes a regression in 3.285.0. Since then the first frame of a multi-frame websocket message was handled as if it were the final frame when using Jetty.

I discovered this when finally getting around to upgrading from 3.284.0 to 4.8.0.0, somewhat unbelievably! 🤯


daviddenton:
Thanks! Releasing now in 4.9.1.0 :)

PR #624, opened 2021-05-31
trimSlashes should not be accessible outside UriTemplate

PaulienVa:
In `UriTemplate` the function `trimSlashes()` is defined as a public extension on String. However, it should only be accessible inside `UriTemplate`


daviddenton:
The wonder of ktlint... 🙄

s4nchez:
That shouldn't be a problem anymore. Thank you for the PR, @PaulienVa

PR #628, opened 2021-07-11
Add digest auth filters

oharaandrew314:
Implemented Client and Server filters for digest auth, excluding widely unsupported features such as the `auth-int` QoP and the `MD5-sess` algorithm.


daviddenton:
thanks! I think this is meaty enough that it merit's it's own module, so I'm going to do a bit of re-orging and then merge it into a branch before tweaking and pulling into master.

PR #629, opened 2021-07-11
Digest auth part2: refactor for style

daviddenton:
Migrated http4k-security into it's own module set and moved the digest auth stuff into http4k-security-digest. Refactored for http4k style and to reuse pre-existing http4k code/concepts (eg. Nonce).

@oharaandrew314 - can you take a quick look before I merge it into master? I think we may be missing some sanity tests around the digest calculations - thought you might be able to advise on that. If you have a working digest-auth setup, it might be worth just trying this implementation out as well before we merge it. Thanks again for your help with this! 😄  (Ignore the formatting changes to the core Server/ClientFilter files - I just moved the DigestAuth stuff out to it's own files. 😉


http4k-security/digest/src/main/kotlin/org/http4k/security/digest/DigestEncoder.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.http4k.core.Method
+import org.http4k.security.Nonce
+import org.http4k.security.digest.Qop.Auth
+import org.http4k.security.digest.Qop.AuthInt
+import org.http4k.util.Hex
+import java.nio.charset.Charset
+import java.security.MessageDigest
+import kotlin.text.Charsets.ISO_8859_1
+
+class DigestEncoder(private val digester: MessageDigest, private val charset: Charset = ISO_8859_1) {
```
oharaandrew314, line 12: Hah!  I had originally called it this, but ended up changing it to `DigestCalculator`.  I'm not opinionated on what we call it.

http4k-security/digest/src/main/kotlin/org/http4k/security/digest/DigestChallenge.kt
```diff
@@ -0,0 +1,45 @@
...
+            .appendIf({ qop.isNotEmpty() }, ", qop=\"$qopValue\"")
+            .appendIfPresent(opaque, ", opaque=\"$opaque\"")
+            .toString()
+    }
+
+    companion object {
+        fun parse(headerValue: String): DigestChallenge? {
+            val (prefix, parameters) = ParameterizedHeader.parse(headerValue)
+
+            return when {
```
oharaandrew314, line 31: Isn't this more complex than short-circuiting with an early return?

oharaandrew314 (commented):
I've been testing the client locally with an IOT device of mine; these new changes work fine with it.

I'm not sure I want to include sample challenges and responses from the device into this repo.  Would you be satisfied with some tests on the `DigestEncoder` using fake data?

oharaandrew314:
I've been testing the client locally with an IOT device of mine; these new changes work fine with it.

I'm not sure I want to include sample challenges and responses from the device into this repo. Would you be satisfied with some tests on the DigestEncoder using fake data?

http4k-security/digest/src/main/kotlin/org/http4k/security/digest/DigestChallenge.kt
daviddenton: Potentially - overall the project style is quite functional - one of expressions over multiple returns. If it's too much then happy to revert.

As for the fake data - yep - totally happy with that. It's the coverage that am most interested in. 🙃

oharaandrew314: I can't commit to this branch, can I?  Shall I embed some new tests as a comment?

daviddenton: You can pull the changes and then open a PR to this branch.

oharaandrew314: Sure, makes sense; improvements are in #630

oharaandrew314 (approved):
Everything seems good to me!  :shipit:   I like the http4k-ifying improvements you've made.

We may want to add some sort of verifiable nonce generator in the near future though; secure random is susceptible to replay attacks from a MITM attack.  Ktor had some sort of encrypted nonce where you could decrypt the expiry date; that seems promising, but requires the user to supply an encryption key.  You can also cache the random nonces, but those need to be evicted, and may not work behind a load balancer depending on the cache implementation.

Alternatively, we could just leave it up to the user if they want to implement their own verifiable nonce generator 🤷 .

daviddenton:
Agreed - I think we can stick with the secure random for now as it's good for the vast majority of cases. If people have more stringent requirements then we've given then the facility to implement.

daviddenton:
Will get thinga moving soon for a release. Thanks again for your help @oharaandrew314  😀

oharaandrew314:
Happy to help!  I'm looking forward to being able to integrate the new module, and delete the ugly hack I made before this.

daviddenton:
New release publishing to MC now... 🙃

PR #635, opened 2021-07-18
be resilient to extra space in Digest challenge header value

oharaandrew314:
Ran into this issue while trying to adapt the client filter for authorization over RTSP.  Still good to be resilient to minor issues like this in any case.


daviddenton:
Thanks 🙃

PR #647, opened 2021-08-10
Downgrade graphql-java and fix Graphql reference example

razvn:
Downgrade graphql-java to 16.2 for compatibility with graphql-kotlin-schema-generator 4.1.1

Wait for 5.0.0 of graphql-kotlin-schema-generator or if they release a fix upgrade to 4.1.1 before going back to 17.0


daviddenton:
Grrr... github actions is down so the release of 4.11.0.1 isn't going through... 🙄

PR #649, opened 2021-08-14
add optional address parameter to Undertow server config

pabl0rg:
sometimes it's necessary to restrict the ip that the server runs on (for example to 127.0.0.1) when using a proxy to reach the internet.

### DIFF
--- a/http4k-server/undertow/src/main/kotlin/org/http4k/server/Undertow.kt
+++ b/http4k-server/undertow/src/main/kotlin/org/http4k/server/Undertow.kt
@@ -10,9 +10,10 @@ import org.http4k.core.HttpHandler
 import org.http4k.sse.SseHandler
 import org.http4k.websocket.WsHandler
 import java.net.InetSocketAddress
+import java.net.InetAddress
 
-class Undertow(val port: Int = 8000, val enableHttp2: Boolean) : PolyServerConfig {
-    constructor(port: Int = 8000) : this(port, false)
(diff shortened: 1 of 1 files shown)


james-richardson-zuhlke:
I'm a bit confused, as this doesn't affect how outbound requests will use a proxy to reach the Internet.
If you mean using a proxy from the Internet to the server, this will only affect the case where the proxy runs on the same host as the java server, otherwise binding to localhost will stop any external proxy reaching the server.
In that case it might be just as easy to use iptables to block external connections, while still listening on all interfaces.

time4tea:
I'm a bit confused as this doesn't affect requests made from the server to the Internet.
In the case where there is a reverse proxy between the Internet and the kotlin server, this change would only make a difference if the proxy was running on the same host, as listening only on loopback would stop an external proxy connecting. In this case would it be easier to use iptables (or similar) to stop connections on non loopback interfaces?

daviddenton:
We have deliberately avoided adding configuration options to the ServerConfigs as we don't want to get into the situation of providing some large amount of options that keeps changing with the underlying backend library, and TBH I'm not very keen to add more to the Undertow one - we already broke our own rule when adding the http2 flag which was probably a mistake. 😄 

The general pattern is that we try to provide reusable chunks so that people can easily create their own custom ServerConfig object if they need to tweak things.

pabl0rg:
Ok that makes sense. its just that the ApacheServer has it so I thought you may have wanted to support it on Undertow.

Thanks

daviddenton:
Thanks anyway! (We don't really like turning down PRs btw 🙃)

PR #655, opened 2021-09-02
add OAuth offline authorization filter

### LINKED ISSUE #654: OAuth Security Client Filter for offline access
I had a use-case for OAuth offline access, where you have a `refresh_token`, `client_id`, and optional `client_secret`, and you must use them to obtain an `access_token` for your requests to the resource server.

I couldn't find anything in the existing OAuth filters to accomplish that, so I made my own.  Was I not searching hard enough?  or would you be interested in me contributing my solution?

oharaandrew314:
Closes #654 

This PR adds an OAuth Offline authorization filter; allowing you to authorize requests to a resource server with the `refresh_token` grant, using the refresh token you saved when your user logged in.

I borrowed a few existing data classes in the `oauth.server` package, and moved them to a new `oauth.core` package, so there's lots of package changes; hope that's ok.  (I've since moved an unused class back to its original spot, which has cut down on the chrun dramatically).

Updated the OAuth reference guide with some info and an example; let me know if I missed anywhere docs are needed.


daviddenton:
Thanks so much for the PR! We will definitely merge this in when we get a chance to look properly... just a bit snowed at the moment. There's definitely a couple small things to tweak (like a typealias to the moved RefreshToken), but happy to take care of that post-merge. 😄

Issue #676, opened 2021-11-18
Regex using greedy qualifier for N times is quoted incorrectly


ruir6o:
```groovy
implementation 'org.http4k:http4k-core:4.17.0.0'
implementation 'org.jetbrains.kotlin:kotlin-stdlib:1.6.0'
```

----

I have just found out about `http4k` and I decided to create a simple service with it. My very first attempt at getting a path parameter to be validated against a UUID RegEx is failing however. This is how I've defined the RegEx:

```kotlin
const val COMPACT_UUID_REGEX = "[a-fA-F0-9]\\{32\\}"
const val PRETTY_UUID_REGEX = "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
const val UUID_REGEX = "(?:${COMPACT_UUID_REGEX}|${PRETTY_UUID_REGEX})"
```

And it is put in a router by using

```kotlin
routes(
    "api/v1/sample/{id:$UUID_REGEX}" bind POST to { /* ... */}
)
```

However running the application yields:
```
Exception in thread "main" java.util.regex.PatternSyntaxException: Unclosed counted closure near index 35
\Qapi/v1/sample/\E((?:[a-fA-F0-9]{32)\Q|[a-fA-F0-9]\E([^/]+)\Q-[a-fA-F0-9]\E([^/]+)\Q-[a-fA-F0-9]\E([^/]+)\Q-[a-fA-F0-9]\E([^/]+)\Q-[a-fA-F0-9]\E([^/]+)\Q)}\E
```

Looking at the exception it seems like `http4k` is replacing the `{` and `}` in an attempt to quote literal path segments, however it is applying it to the regex section as well, interfering with the greedy qualifiers defined in the UUID RegEx.

daviddenton:
Thanks for the report. It looks like a tricky one to solve TBH (Regexes 🙄 ). I've added a failing test anyway [here](https://github.com/http4k/http4k/blob/c509eaa5baf3c090bceaa4cba428396ed1fb2c11/http4k-core/src/test/kotlin/org/http4k/core/UriTemplateTest.kt#L138)

ruir6o:
@daviddenton I think a simple custom parser could do the trick. This is just something I've whipped up in a few minutes to use as an example of how a parser could help:

```kotlin
fun parsePath(path: String): List<Token> {
    val chars = path.chars().iterator()
[...]

PR #678, opened 2021-11-28
Add URI parser

ruir6o:
Replaces the UriTemplate logic by a custom parser that handles complex regular expressions (i.e. regular expressions that contain the delimiters for the path parameters).

The focus is on keeping the public API of the `UriTemplate` 100% compatible with the previous implementation while overhauling its internal handling of regular expressions in path parameters.

This PR is mostly raised as an example of a potential fix for #676, it does not attempt to include everything that is required in terms of quality checks for an actual contribution, although it can be added if needed.


javadev:
This branch has conflicts that must be resolved

s4nchez:
This PR has been open for two years now and it's safe to say it's not being merged.

As much as we'd like to have a better UriTemplate, I don't think we can justify maintaining a full-blown parser implementation at this stage.

PR #688, opened 2022-01-14
The woff2 font format is now widely supported. https://caniuse.com/woff2

simon-broadley:
I can't see a reason why the woff2 format wouldn't be supported out-of-the-box. It's widely supported and has much better compression than woff. See browser support here: https://caniuse.com/woff2


daviddenton:
Thanks! It wasn't a deliberate snub 😉

Issue #602, opened 2021-03-24
Incorrect use of URLDecoder in UriTemplate


vojtechtoman:
The use of `URLDecoder` in: https://github.com/http4k/http4k/blob/791907b1662b78b1812e4ec21e7bfc823808c1e1/http4k-core/src/main/kotlin/org/http4k/core/UriTemplate.kt#L38 leads to incorrect results because we are in the 'path' URI component where different escaping rules apply compared to the 'query' URI component. For instance, the '+' character should be treated literally in a URI path, but `URLDecoder` replaces it with a space.

s4nchez:
@vojtechtoman, thank you for reporting this. 

Can you please provide a test showing how the problem manifests and some reference (maybe from the RFC?) on how the correct behaviour should be? That'd speed up any potential fix as that code is from the early days of the project, and I need to re-familiarize myself with that part of the domain.

vojtechtoman:
The relevant part of [RFC 3986](https://tools.ietf.org/html/rfc3986) are the `segment` and `pchar` [productions](https://tools.ietf.org/html/rfc3986#appendix-A):
```
segment       = *pchar
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
pct-encoded   = "%" HEXDIG HEXDIG
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
              / "*" / "+" / "," / ";" / "="
```
The characters ":", "@", `unreserved`, and `sub-delims` have no special meaning in a path segment and should not be interpreted.

As for a reproducer:
```
val routes = routes(
    "/test/{param}" bind GET to {
        println(it.path("param"))
        Response(Status.OK)
    }
)
routes(Request(GET, "/test/a+b")) // outputs "a b" instead of "a+b"
```

mrtom808:
The path encoder also looks incorrect as it encodes " " to "+" instead of "%20". URLDecoder/URLEncoder are for working with form data encoded after a ? in a URL so their use in UriTemplate for paths should be avoided. Unfortunately the jdk libs hide anything useful that could be used.

**UriTemplateTest**:
```java
    @Test
    fun supportsPlusCharactersInUriPath() {
        val template = from("properties/{name}")
        val parameters = template.extract("properties/A%20+%20B")
        assertThat(parameters.getValue("name"), equalTo("A + B"))
[...]

PR #690, opened 2022-01-17
Fixed URLPathSegment encoding/decoding based on RFC 3986

jenarros:
See [appendix A (https://www.ietf.org/rfc/rfc3986.txt)](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A) for details about valid path segments. In summary the spec indicates that a valid path segment can contain:
```
segment       = *pchar
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
pct-encoded   = "%" HEXDIG HEXDIG
unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
              / "*" / "+" / "," / ";" / "="
```

The implementation can probably be optimised but the PathTest.kt tests should now capture the specification more accurately and `toPathSegmentEncoded` / `toPathSegmentDecoded` are used to encode/decode path segments.

This will change the current path encoding behaviour allowing the characters above, although technically it is a fix, I am not sure whether it should be flagged as a breaking change.

Fixes http4k/http4k#602

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRoute.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/ContractRoute.kt
@@ -14,7 +14,7 @@ import org.http4k.core.Status.Companion.NOT_FOUND
 import org.http4k.core.Status.Companion.OK
 import org.http4k.core.Uri
 import org.http4k.core.then
-import org.http4k.core.toPathDecoded
+import org.http4k.core.toPathSegmentDecoded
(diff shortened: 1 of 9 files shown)


daviddenton:
Thanks VM for your efforts on this @jenarros . Since it is a change in behaviour we're going to have to take a closer look at this OFC. 

We'd definitely tag it as a breaking change anyway, just to be sure and to be the least surprising to userbase. 😄

PR #692, opened 2022-01-21
fix: Keep-alive for Netty when not streaming

jakubjanecek:
I have noticed that the Netty implementation does not respect connection keep-alive and the reason is because `HttpServerKeepAliveHandler` from Netty disables keep-alive whenever the content-length is unknown: https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerKeepAliveHandler.java#L115

So I reimplemented `Http4kChannelHandler` to differentiate between the streaming and non-streaming case.

### DIFF
--- a/http4k-core/src/test/kotlin/org/http4k/server/ServerContract.kt
+++ b/http4k-core/src/test/kotlin/org/http4k/server/ServerContract.kt
@@ -194,6 +194,18 @@ abstract class ServerContract(private val serverConfig: (Int) -> ServerConfig, p
             ))
     }
 
+    @Test
+    fun `sets keep-alive for non-streaming response`() {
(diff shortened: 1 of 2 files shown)


daviddenton:
Thanks!

jakubjanecek:
@daviddenton Just wondering (so that I understand the next time), why did the build fail even though it worked on my machine and I am quite sure it was supposed to work in CI too?

daviddenton:
It’s interesting that the build passed on your machine, because the contract actually failed for a few of the implementations. For the moment I’ve moved the test from the ServerContract down to the Netty implementation. We can look at making the others pass and then pulling it back up again.

:)

jakubjanecek:
Oh, alright, I only ran the tests of the Netty module. Makes sense now.

PR #710, opened 2022-03-11
Fix Strikt assertion builder for Uri.path

### LINKED ISSUE #709: Strikt assertion builder on Uri.path gets host instead
Using http4k-testing-strikt, this test should pass:
```kotlin
expectThat(Uri.of("http://bob/bill")).path isEqualTo "/bill"
```
but it fails with
```
▼ Expect that http://bob.com/bill:
  ▼ value of property host:
    ✗ is equal to "/bill"
            found "bob"
```

michaelbannister:
Fixes #709


daviddenton:
Thanks!

PR #712, opened 2022-03-19
add case-insensitive enum lenses for path, query, and header

### LINKED ISSUE #708: Case insensitivity for enum lenses
There are time I would like an enum lens be case-insensitive.  Are there any objections to adding this as an option?

e.g.
```
val queryLens = Query.enum<MyType>(caseSensitive=false).required("type")
val pathLens = Path.enum<MyType>(caseSensitive=false).of("type")
```

I wouldn't mind contributing this feature.

oharaandrew314:
Closes #708


daviddenton:
Thanks!

PR #715, opened 2022-04-01
add disk cache path to MultipartFormBody.from() parameters

rny:
Working as server to handle file upload

Now the disk cache can only be cleaned using deleteOnExit(). That means keep all the temp files until restart. But server is normally keep running without restart. 

The pull is simply put the disk cache path into parameters so we can provide the cache path and clean the cache ourselves when each request is finished.


daviddenton:
The build for this failed. If we can get it working then happy to look at merging it 😄

rny:
Forget the import, it is fixed now

PR #717, opened 2022-04-02
Make CatchAll filter configurable to not expose stacktrace to clients

dmcg:
It's a breaking change if people were relying on the stack trace, but they can largely opt back in to the original version.


daviddenton:
Hmm. i think we should keep the original behaviour as the defaultc whilst giving people the option to override. And the dodgy comment shoukd go as well. 😉

dmcg:
It’s your call. The default is a little dodgy, and will be what is used out of the box, but backward compatibility is good. 

Ultimately I’d like to make the default talk to Events

daviddenton:
I think there is an entire module to be had around the events ecosystem - plugging it into structured logging, the tracerbullet stuff, metrics etc...

PR #721, opened 2022-04-09
Add zoneId, zoneOffset, and locale as default mappings

### LINKED ISSUE #720: Additional Standard Mappings
I'd like to propose additional standard mappings to `AutoMappingConfiguration`.

- java.util.Locale
- java.time.ZoneId
- java.time.ZoneOffset

I can follow up with a PR if accepted.  Thoughts?

oharaandrew314:
Closes #720

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/lens/BiDiMapping.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/lens/BiDiMapping.kt
@@ -19,6 +19,8 @@ import java.time.LocalTime
 import java.time.OffsetDateTime
 import java.time.OffsetTime
 import java.time.YearMonth
+import java.time.ZoneId
+import java.time.ZoneOffset
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeFormatter.ISO_INSTANT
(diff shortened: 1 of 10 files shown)


daviddenton:
👍

daviddenton:
@oharaandrew314 We've just upgraded in a project to the version with this change and there seems to be a problem with the mapping of Locale - we're not sure if mapping a locale to a language tag is the correct thing to be doing. Can you provide more details about this?

oharaandrew314:
Are you able to share some details about the error?

daviddenton:
sure - we're deserialising "pl_PL" - which is a supported locale as per this: https://www.localeplanet.com/icu/pl-PL/index.html

The new mapping causes this value to be deserialised as a blank string.

oharaandrew314:
The java locale on that page should be `pl-PL`, but I'll see if I can get it to support the hyphenated form.

daviddenton:
that's the strange thing - presumably the wire format should not be the java version? Although TBF I think the -/_ is irrelevant. Not sure! 😂

oharaandrew314:
@daviddenton I did some research, and it seems like the hyphenated format is the [ISO ](https://en.wikipedia.org/wiki/ISO_639_macrolanguage) standard, and the underscored format is  [IETF ](https://en.wikipedia.org/wiki/IETF_language_tag) standard.  Java can only parse the IETF standard, so that's why the hyphenated locale cannot be parsed.

Apparently supporting both isn't as simple as replacing the `-` with a `_` because there are corner cases, and even if we added apache-commons `LangUtils` as a core dependency to parse the ISO format--doubt we want that--then we would still have to choose just one format to serialize to.

I think we should leave this as-is.  Although, perhaps a lens failure should be thrown rather than return an empty locale.

oharaandrew314:
Sorry for all the edits; I keep getting the hyphen and underscore mixed up.

PR #723, opened 2022-04-14
Fix header rendering for effectively empty directive values for FallbackCacheControl

matsluni:
This PR removes empty directive values in FallbackCacheControl.

I noticed this when testing this in my codebase with zero durations for `stale-while-revalidate` and `stale-if-error`, that the `Cache-Control` header value generates something like `public, max-age=10, , `.


daviddenton:
Thanks! 🙃

PR #724, opened 2022-04-15
Throw a lens failure if a valid locale was not parsed

oharaandrew314:
Relates to #721 

Java locales cannot parse ISO language codes, but will happily return an empty `Locale` object, which can lead to confusion.  If accepted, this lens validation will only accept locales where a language was detected, and this updated test will formalize the notion that ISO format language codes are not supported.


daviddenton:
👍

PR #729, opened 2022-04-21
Fix #728: Add requestModifier param to JavaHttpClient

### LINKED ISSUE #728: No way to set the request timeout when using the JavaHttpClient
Using the `JavaHttpClient.invoke()` function, it is possible to inject your own custom made `JavaHttpClient`. This allows you to set a connection timeout, ssl context, a proxy, etc. However, for the `JavaHttpClient`, the request timeout is set on the request. There is no way to configure this.

I would like to be able to configure the request timeout for the `JavaHttpClient`.

As I see it, there are several solutions:
- Since it is a request timeout: add it to the request. It seems to me that this solution is not on the table.
- Add an optional parameter to the `JavaHttpClient.invoke()` funtion that specifies the request timeout
- Add a parameter to the `JavaHttpClient.invoke()` funtion that is a receiver function on the `HttpRequest.Builder`, with as a default the no-op function. This allows the user to modify their requests.

gmulders:
Note that, although I suggested a receiver function in issue 728, I noticed that the JettyClient had a similar solution using a function `(JettyRequest) -> JettyRequest`, so I followed that example.

Also note that I enabled the `socket timeouts are converted into 504` test-function.

### DIFF
--- a/http4k-client/fuel/src/test/kotlin/org/http4k/client/FuelStreamingTest.kt
+++ b/http4k-client/fuel/src/test/kotlin/org/http4k/client/FuelStreamingTest.kt
@@ -4,8 +4,13 @@ import org.http4k.core.BodyMode.Stream
 import org.http4k.server.ApacheServer
 import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
+import java.time.Duration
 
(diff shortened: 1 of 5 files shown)


daviddenton:
Thanks!

Issue #735, opened 2022-05-08
Etag filter doesn't handle StreamBody


aSemy:
I'm using http4k to host some images. I added the Etag filter, but I noticed it always returns the same header value.

I debugged `object AddETag` and noticed that it's doing `response.body.toString()`. This returns the string `<<stream>>`, and the MD5 of that will always be the same - `5459d1b8f819085b9c0a179c8c292dbe`

https://github.com/http4k/http4k/blob/0508600fef8d84891cb8dec343860823fb76da90/http4k-core/src/main/kotlin/org/http4k/filter/CachingFilters.kt#L101-L113

Might replacing `response.body.toString()` with `response.bodyString()` work instead?

daviddenton:
That does look incorrect. It's not using the content of the body either - just the object identity of the body object.

As for the stream implementation - you obviously can't really use an etag without pulling the entire stream into memory, so I'm not sure what we would want to do there. Maybe as a trailer (which would involve us adding trailers 🙃)

PR #736, opened 2022-05-08
#735 use whole message body for etag hash

aSemy:
Resolves https://github.com/http4k/http4k/issues/735?

hi @daviddenton, here's a quick fix for #735. I'm not sure what you mean by trailers, and as you note the performance will be bad for large response bodies, so I can understand why this shouldn't be merged. I'll leave it as a draft.

Anyway - at least the tests are useful!


daviddenton:
thanks!

Issue #738, opened 2022-05-16
Etag filter is broken: prevents content from being sent from the server


stevebakh:
The Etag filter was previously broken, using the response from `toString` to generate the hash.

An attempt to fix this was pushed in commit: 4f1491f5ad2cdd09b1e7a8d94ab97d3f581966bf

A subsequent commit simplified the code (but left it still in a broken state): 5a8d989859bfa4e863b4adc534dbea967474387d - reading the byte array in this way moves the pointer and results in the HTTP response containing a valid Etag, but a 0 length payload!

daviddenton:
Whoop! Thanks for letting us know. We'll fix it

stevebakh:
I don't think the proposed change fixes the issue. The problem is that the input stream is being read and the pointer not reset (and looking at the code, these input streams don't support reset).

I'm guessing there are no automated tests around this functionality, as the code will return a 0-length payload in the HTTP response (with a correct Etag at least)?

daviddenton:
We could read the entire content into memory and then write it back out again in the filter chain response?

stevebakh:
If it helps, this is the cheap workaround I've put in place to test something, and it works as expected, but I'm not sure how performant this solution is:

```
val payload = response.body.payload.array()
val hashedBody = md5().digest(payload).joinToString("") { "%02x".format(it) }
response.header("Etag", hashedBody).body(ByteArrayInputStream(payload))
```

Note the need to set the response body again at the end. :(

daviddenton:
With a stream there isn't much option really - we need the entire thing in memory so that we can calculate the tag

If you've got time to PR it then we'll happily merge :)

jshiell:
Please, don't inspect streams. This is _hugely_ risky for any large stream, as it'll force it into memory - and that's **bad** for some of our streams.

I don't believe there's any generic way to generate an eTag for streamed content - it'll need to be done by the application in a context specific manner (e.g. getting it from an upstream, or storing the etag somewhere for retrieval at delivery time).

daviddenton:
so what is our preferred option for a stream? We either 1. do nothing - which seems weird, 2. throw up, 3. JFDI and assume people know not to add the filter to streams 😄 

I'm not sure there is a good solution really.

jshiell:
I think (1) is correct, as you want the filter to be applied to in-memory bodies, and you don't want it to break streams. Log a warning if you must. 

[...]

PR #739, opened 2022-05-16
Drop AddETag in favour of more fully-featured ETagSupport filter

jshiell:
As discussed in #738 , this drops the `AddETag` filter in favour of `ETagSupport`, which properly skips streaming responses and also handles a 304 response where appropriate.

This does leave us without a standard filter to provide _only_ etag headers, but that seems a very niche case, and it's definitely worth reducing the confusion here.


jshiell:
You're no fun 😛

PR #740, opened 2022-05-17
Support parsing digest challenge where nonce has an '='

oharaandrew314:
I don't know why some devices think they can put an `=` in a nonce, but this fix will allow us to correctly parse the challenge.


s4nchez:
Good spot. Thank you for the fix, @oharaandrew314

PR #741, opened 2022-05-22
Add `queryParametersEncoded` + query param value split bug fix

overfullstack:
- Add `queryParametersEncoded` extension to Uri to encode only parameter values
- :bug: Change `split("=")` to `split("=", limit=2)` to deal with parameter values that contain `=`


s4nchez:
Thank you for spotting and fixing this, @overfullstack !

PR #746, opened 2022-06-15
For Discussion: Moshi AutoMarshallingJson

oharaandrew314:
I wanted to add `AutoMarshallingJson` capabilities to Moshi, so I could use it for OpenApi spec generation in my lightweight AWS Lambda APIs.  Presumably, the reason this was never done is because Moshi only supports object mapping and json streaming, with no intermediary `NODE` format, as required by `AutoMarshallingJson<NODE>`.  I came up with my own `MoshiElement` format, which went pretty well at first.

As, you'll see in `ConfigurableMoshi`, I didn't have a great way of implementing `fun asJsonObject(input: Any): MoshiElement` and `fun <T : Any> asA(j: MoshiElement, target: KClass<T>): T`, so I used the existing Moshi `JsonAdapter` capabilities to convert the object to Json before reading to the `NODE` format.  This isn't ideal, but I think implementing it properly would involve performing custom reflection on the classes, which would overlap with Moshi's `JsonAdapter` responsibilities.

Unfortunately, when I try to use this to generate my OpenApi Spec, `fun asJsonObject(input: Any): MoshiElement` is called to convert `org.http4k.contract.openapi.v3.ArrayItems`, and when that's delegated to `asFormatString`, the conversion to json fails because Moshi can't generate an `JsonAdapter` for `ArrayItems` (an interface).

```kotlin
OpenApi3(
    ApiInfo(title = "My API", version = "v1.0"),
    json = Moshi,
    apiRenderer = ApiRenderer.Auto(Moshi, AutoJsonToJsonSchema(Moshi))
)
```

```
java.lang.IllegalArgumentException: No JsonAdapter for interface org.http4k.contract.openapi.v3.ArrayItems (with no annotations)
for interface org.http4k.contract.openapi.v3.ArrayItems items
for class org.http4k.contract.openapi.v3.SchemaNode$Array
	at com.squareup.moshi.Moshi$LookupChain.exceptionWithLookupStack(Moshi.java:389)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:158)
	at com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory.create(KotlinJsonAdapter.kt:292)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:146)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:106)
	at com.squareup.moshi.Moshi.adapter(Moshi.java:80)
	at org.http4k.format.ConfigurableMoshi.asFormatString(ConfigurableMoshi.kt:34)
```

**tl;dr**
- No way to efficiently convert from `Any->MoshiElement` and `MoshiElement->T`
- Moshi's adapter format can't support writing the openapi models without custom adapters

### DIFF
--- a/http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiElement.kt
+++ b/http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiElement.kt
@@ -1,0 +1,11 @@
+package org.http4k.format
+
+sealed interface MoshiElement
+
+data class MoshiArray(val elements: List<MoshiElement>): MoshiElement
+data class MoshiObject(val attributes: Map<String, MoshiElement>): MoshiElement
+sealed interface MoshiPrimitive: MoshiElement
+data class MoshiString(val value: String): MoshiPrimitive
+data class MoshiNumber(val value: Number): MoshiPrimitive
+data class MoshiBoolean(val value: Boolean): MoshiPrimitive
+object MoshiNull: MoshiPrimitive
--- a/http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiElement.kt
+++ b/http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiElement.kt
@@ -1,6 +1,14 @@
 package org.http4k.format
 
-sealed interface MoshiElement
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import okio.Buffer
+import okio.buffer
+import okio.source
+
+sealed interface MoshiElement {
+    companion object
+}
 
 data class MoshiArray(val elements: List<MoshiElement>): MoshiElement
 data class MoshiObject(val attributes: Map<String, MoshiElement>): MoshiElement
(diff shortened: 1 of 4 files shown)


http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiElement.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+private fun JsonReader.parse(): MoshiElement {
+    return when(val token = peek()) {
+        JsonReader.Token.BEGIN_ARRAY -> parseArray()
+        JsonReader.Token.BEGIN_OBJECT -> parseObject()
+        JsonReader.Token.STRING -> MoshiString(nextString())
+        JsonReader.Token.BOOLEAN -> MoshiBoolean(nextBoolean())
+        JsonReader.Token.NUMBER -> MoshiNumber(nextDouble())
```
oharaandrew314, line 70: Moshi has no way of telling you whether the current number is decimal or whole.  Possible loss of precision here.

daviddenton:
I assume you're attempting to do it without reflection here. In that case, the various adapters will need to be written, or generated with something like Kotshi. 

As for the interface lookup, you can use IsAnInstanceOfAdapter to convert the lookup to one that is explicitly registered for the class in question.

oharaandrew314:
David, so you think we should go all the way to reinvent the wheel and try to make our own algorithms for `T->NODE` and `NODE->T` mapping?

PR #748, opened 2022-06-16
Refreshing Credentials Provider test improvement

makowalski:
There is an edge case where the calls for second refresh and chasing refresh are already done before test asserts counter for only second refresh.

Co-authored-by: mandyvuong <mandyvuong@gmail.com>


daviddenton:
😄

PR #749, opened 2022-06-22
Fix multipart upload failure if charset is included in content type

wickwirew:
If the charset was included in the `Content-Type` header then it would end up using that as the boundary instead of the actual defined boundary.

This is my first contribution to http4k so if there is something I missed let me know!


s4nchez:
Good spotting, @wickwirew. Thank you for your contribution!

PR #752, opened 2022-07-08
Build logic improvements

c00ler:
This Pull Request proposes small improvements to the build script logic to improve performance and maintainability.

Changes:
* Remove `openapi-generator-gradle-plugin`
* Migrate `http4k-contract` build script to Kotlin DSL to be consistent with the rest of the modules

I was looking at the configuration of `org.openapitools.generator.gradle.plugin.tasks.GenerateTask` tasks. The way they are configured makes them not cacheable https://scans.gradle.com/s/cnwrkzmkwogak/timeline?cacheability=overlapping_outputs. They write the output into the same directory, therefore gradle is not able to infer the cacheability of these tasks. Also, the output folder is actually located outside of the gradle workspace and generated files are not included in the build. Unless I'm missing something, it looks to me that these tasks can be removed from the build. Removal of tasks makes the build logic a little bit more straightforward, and faster, and requires fewer dependencies.


daviddenton:
The tasks that have been removed are used as a part of our build process to check compatibility against the open api generator, so can't be removed I'm afraid.

Issue #753, opened 2022-07-14
GraphQL explorer for http4k-graphql


arnabkd:
**GraphQL explorer endpoint**

It would be nice to have a graphical IDE for developers to try out there code and explore the graphql endpoint one creates. [ktor-graphql](https://github.com/excitement-engineer/ktor-graphql) has this functionality, which can be used like so:

```
val server = embeddedServer(Netty, port = 8080) {
    routing {
        graphQL("/graphql", schema) { request ->
            Config(showExplorer = true)
        }
    }
}
...
```

I get that http4k is simplistic and minimalist at it's very core, and I respect that. If this feature contradicts that, feel free to close. I do feel that this will provide some very handy functionality :)

razvn:
We just add a static endpoint with [GraphQL-playground](https://github.com/graphql/graphql-playground) html page. 
Not sure it needs to be packed by the library.

daviddenton:
It would be trivial to make a reusable route module which serves a static HTML file from the resources folder and contains the playground content. It won't even add more dependencies so we can just ship it with the existing graphql module

arnabkd:
> We just add a static endpoint with [GraphQL-playground](https://github.com/graphql/graphql-playground) html page. Not sure it needs to be packed by the library.

Not really a question of "need" per se, as this can always be resolved by putting the playground in the frontend or even by using a standalone graphql ide. It's more a question of convenience.

daviddenton:
If someone wants to PR this into the graphql module with no additional.dependencies then we can ship it 🙃

arnabkd:
@daviddenton : Added a PR now. This is my first time contributing to http4k, so I am unsure what the expectations are :)

daviddenton:
Fixed for next release.

PR #754, opened 2022-07-18
GraphQL playground

arnabkd:
This PR adds a static graphql playground on a configurable route function.

See: https://github.com/http4k/http4k/issues/753


arnabkd:
I really wanted to extend the existing `graphql` functions to incorporate the playground, but I am unsure how to do that and also I think that might add a bunch of extra arguments to the existing functions

arnabkd:
> I think it's a good idea to keep it separate. One of the nice things about the http4k model is that things are pretty composable so you can mix and match things. 🙃

Yes, I thought this solution was probably more in line with how http4k is structured. Thanks for the review :)

PR #755, opened 2022-07-20
Make Moshi an AutoMarshallingJson

oharaandrew314:
I wasn't able to make the `NODE` type an `Any`, despite moshi allowing you to work with it as in intermediate type.  Eventually you run into an issue where your `NODE` has to hold a `null`, which isn't allowed by the type contract.


http4k-format/moshi/src/test/kotlin/org/http4k/format/MoshiTest.kt
```diff
@@ -42,9 +44,128 @@ class MoshiAutoTest : AutoMarshallingJsonContract(Moshi) {
...
+                "null" to nullNode(),
+                "int" to number(2),
+                "empty" to obj(),
+                "array" to array(listOf(
+                    string(""),
+                    number(123)
+                )),
+                "singletonArray" to array(obj("number" to number(123)))
+            )
+            val expected = """{"string":"value","double":1.5,"long":10,"boolean":true,"bigDec":1.2,"bigInt":12344,"int":2,"empty":{},"array":["",123],"singletonArray":[{"number":123}]}"""
```
oharaandrew314, line 167: Difference here is that the `bigDec` doesn't suffer from the floating-point imprecision of the original, and `null` is missing, because moshi doesn't serialize nulls, by default.

http4k-format/moshi/src/main/kotlin/org/http4k/format/MoshiNode.kt
```diff
@@ -0,0 +1,50 @@
...
+    null -> MoshiNull
+    is Iterable<*> -> obj
+        .map { wrap(it) }
+        .toList()
+        .let { MoshiArray(it) }
+    is Map<*, *> -> obj
+        .mapKeys { (key, _) -> key.toString() }
+        .mapValues { (_, value) -> wrap(value) }
+        .let { MoshiObject(it) }
+    is Number -> {
```
oharaandrew314, line 39: Not sure sure about this, because it would convert an intentional `123.0` to an integer.  But with the way Moshi works, it's either this, or convert all numbers to decimal.

http4k-format/moshi/src/main/kotlin/org/http4k/format/ConfigurableMoshi.kt
```diff
@@ -95,8 +104,11 @@ fun Moshi.Builder.asConfigurable() = object : AutoMappingConfiguration<Moshi.Bui
         }
 
     // add the Kotlin adapter last, as it will hjiack our custom mappings otherwise
-    override fun done() =
-        this@asConfigurable.add(KotlinJsonAdapterFactory()).add(Unit::class.java, UnitAdapter)
+    override fun done(): Moshi.Builder {
+        adapter(BiDiMapping(Double::toBigDecimal, BigDecimal::toDouble), { value(it) }, { nextDouble() })
```
oharaandrew314, line 108: Since `MoshiNode` holds `BigDecimal` and `BigInteger`, I have to add a low-level adapter for them.  It feels weird, but I feel like it's the only way to be immune to floating-point imprecision.

oharaandrew314:
@daviddenton I've incorporated your changes.  Should be ready for merge.

daviddenton:
Am releasing this now as a part of 4.27.4.0. Thanks!

PR #756, opened 2022-07-22
Do not throw exception when iteration over empty TestWsClient.received sequence

oharaandrew314:
Guarding against this by checking if the sequence `hasNext` is all well and good, but I think we should be able to iterate without fear.  For Example, this would throw `NoSuchElementException` if the client had no messages.

```kotlin
for (message in client.received()) {
  // do stuff
}
```


daviddenton:
👍👍

PR #758, opened 2022-07-23
Add userData map to HttpMessage to smuggle data in and out

dmcg:
I'm not wedded to the implementation, especially the nasty self-type extension functions (well I'm secretly proud of them, but happy to see them replaced if you have a better way), but interested in feedback on the interface. 

In particular - `Map<Any, Any>` feels like it shuts least doors, but `Map<String, Any>` may communicate better, and if people really need a map with arbitrary keys they can add their own to the userData map?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ changes with their rationale when appropriate:
 ### v4.47.0.0 (uncut)
 - **http4k-server-format-kondor-json**: [New Module] Support for [KondorJson](https://github.com/uberto/kondor-json)
 
+### v4.46.0.1
+- **http4k-incubator** Fix TracerBuller to handle non-root spans when building trees
+
 ### v4.46.0.0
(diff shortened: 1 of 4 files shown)


oharaandrew314:
I wonder if [Request Contexts](https://www.http4k.org/guide/howto/attach_context_to_a_request/) might accomplish what you want?  Their goal is to attach metadata to a request and make it available through each step of the filter chain.  For example, you have an authorizer filter that adds the user id to the context, so that it can be retrieved in a future step.

daviddenton:
This is more to do with attaching data to the Response as opposed to the request, as per Duncan's video here:

https://kotlinlang.slack.com/archives/C5AL3AKUY/p1658568583645649

dmcg:
I’d also like to be able to access any such data armed only with the message itself, so that we don’t have to pass context down to any any places that want to get or set it.

PR #762, opened 2022-08-05
add cors exposed headers property

oharaandrew314:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers


daviddenton:
thanks!

PR #764, opened 2022-08-07
Add Auto Content Negotiation

oharaandrew314:
This utility wraps several body lenses simplify content negotiation for inbound and outbound messages.

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/routeMeta.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/routeMeta.kt
@@ -133,6 +133,7 @@ class RouteMetaDsl internal constructor() {
         for (lens in negotiated.first) {
             receiving(lens to negotiated.second, definitionId, schemaPrefix)
         }
+        requestBody = negotiated.first.toBodyLens()
     }
 
     /**
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/routeMeta.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/routeMeta.kt
@@ -8,6 +8,7 @@ import org.http4k.core.Request
 import org.http4k.core.Response
 import org.http4k.core.Status
 import org.http4k.core.with
+import org.http4k.format.ContentNegotiator
 import org.http4k.lens.BiDiBodyLens
 import org.http4k.lens.BodyLens
 import org.http4k.lens.Header
@@ -91,6 +92,23 @@ class RouteMetaDsl internal constructor() {
     }
 
     /**
+     * Add all negotiated responses, with samples.  It is also possible to pass in the definitionId
+     * for this response body which will override the naturally generated one.
+     */
+    @JvmName("returningStatusNegotiated")
+    fun <T: Any> returning(
+        status: Status,
+        negotiated: Pair<ContentNegotiator<T>, T>,
+        description: String? = null,
+        definitionId: String? = null,
+        schemaPrefix: String? = null
+    ) {
+        for (lens in negotiated.first) {
+            returning(status, lens to negotiated.second, description, definitionId, schemaPrefix)
+        }
+    }
+
+    /**
      * Add an example request (using a Lens and a value) to this Route. It is also possible to pass in the definitionId
      * for this request body which will override the naturally generated one.
      */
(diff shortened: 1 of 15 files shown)


http4k-contract/src/main/kotlin/org/http4k/contract/routeMeta.kt
```diff
@@ -102,6 +120,22 @@ class RouteMetaDsl internal constructor() {
...
+    @JvmName("receivingNegotiated")
+    fun <T> receiving(
+        negotiated: Pair<AutoContentNegotiator<T>, T>,
+        definitionId: String? = null,
+        schemaPrefix: String? = null
+    ) {
+        for (lens in negotiated.first) {
+            receiving(lens to negotiated.second, definitionId, schemaPrefix)
+        }
+        requestBody = negotiated.first.toBodyLens()
```
oharaandrew314, line 136: `RouteMeta` can only support a single body lens.  I attempted to convert it to a list, but ran into issues.  This compromise isn't perfect, but seems to work well enough.

http4k-format/core/src/main/kotlin/org/http4k/format/AutoContentNegotiatior.kt
```diff
@@ -0,0 +1,41 @@
...
+        ?.let { contentType -> Accept(listOf(contentType), emptyList()) }
+        .let { accept -> invoke(accept) }
+        .invoke(target)
+
+    fun outbound(request: Request): BiDiBodyLens<T> = invoke(Header.ACCEPT(request))
+
+    fun toBodyLens() = BodyLens(
+        metas = defaultLens.metas + alternateLenses.flatMap { it.metas },
+        contentType = defaultLens.contentType,
+        getLens = this::invoke
```
oharaandrew314, line 33: The `getLens` override will allow the preflight check to use the appropriate lens.  I'm not sure if appending all of the metas is the right thing to do here.

http4k-format/core/src/test/kotlin/org/http4k/format/AutoContentNegotiatorTest.kt
```diff
@@ -0,0 +1,113 @@
...
+            .header("accept", v1Lens.contentType.toHeaderValue())
+            .let(returnJohn)
+
+        assertThat(response, hasBody("v1-john"))
+    }
+
+    @Test
+    fun `response body - with v2 accept header`() {
+        val response = Request(GET, "/")
+            .header("accept", v2Lens.contentType.toHeaderValue())
```
oharaandrew314, line 61: Can the `ACCEPT` lens be used to inject a `Content-Type` into the request?  Seems like it's only outbound.

daviddenton: it is only an extractor at the moment - mostly because we didn't want to get into dealing with the complexity of the directives and q-levels: ```Accept: text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8```

http4k-format/core/src/main/kotlin/org/http4k/format/AutoContentNegotiatior.kt
daviddenton: Metas are used for a couple of things - 

1. to generate error messages when there are things missing or invalid from a request
2. to generate the OpenApi schema for form bodies.

So I think you're ok for this TBH.

src/docs/guide/howto/use_auto_content_negotiation/index.md
```diff
@@ -0,0 +1,38 @@
+title: http4k How-to: Use Auto Content Negotiation
```
daviddenton, line 1: Thanks for writing the docs for this! 😄

daviddenton:
If you think this is ready to go then am happy to merge it. I might do a little bit of tweaking before release, but overall 😄😄😄!

oharaandrew314:
Yeah, if you think this would be a good addition, I think it's ready for merge, and your final touches.  I'd be interested to see if you can get the `AutoContentNegotiator` more in line with http4k's functional style.

daviddenton:
So quick question. Can we think of any reason for this method to be public?
```
    operator fun invoke(accept: Accept?): BiDiBodyLens<T>
        = find { accept == null || accept.accepts(it.contentType) } ?: defaultLens
```

oharaandrew314:
It's actually the most important method.  If any of them deserve to be the face of the `AutoContentNegotiator`, it's that.  The other `invoke` and `oubound` could just be extension methods.  But then again, there's little practical reason for anyone to call it, so it could be private.

Also, I made a mistake in the docs.  This feature depends on `http4k-format-core`, and not `http4k-core`.

daviddenton:
I can fix that up. don’t worry

daviddenton:
Having had a better look, this was actually a really neat implementation - that you so much for contributing it :). The only method that slightly irks me is the outbound() method - I’d prefer somehow for the negotiator to be a function in that way as well, but I also can’t see how to do it!

oharaandrew314:
Thanks!  At first I thought this would be a very simple feature for outbound responses, but then I realized it can be used for inbound, and contract generation.

I'm not super happy with the `outbound` method either, but we absolutely do need to pass the original request or accept header in to select the inner lens, so there's no working around it.  Perhaps some name changes might be in order, or unifying the mixture of operator and non-operator functions.

Issue #763, opened 2022-08-06
Moshi - Replace Kotlin-Reflect with kotlinx-metadata-jvm


oharaandrew314:
Moshi has an [approved PR](https://github.com/square/moshi/pull/1183) that replaces kotlin-reflect with kotlinx-metadata-jvm; this PR was approved 2 years ago, but has sadly been left to rot.

This change would be nice because kotlin-reflect has grown to be bloated, and the much leaner kotlinx-metadata-jvm can get the same job done.  In serverless environments, every Megabyte helps, and this change saves 2.  In the meantime, the author has released [their own adapter](https://github.com/ZacSweers/MoshiX/tree/main/moshi-metadata-reflect), which continues to be maintained.  It's quite easy to integrate into http4k:

```kotlin
implementation("org.http4k:http4k-format-moshi") {
    exclude("org.jetbrains.kotlin", "kotlin-reflect")
}
implementation("dev.zacsweers.moshix:moshi-metadata-reflect:0.18.3")
```

```kotlin
object MoshiLite : ConfigurableMoshi(
    Moshi.Builder()
        .addLast(EventAdapter)
        .addLast(ThrowableAdapter)
        .addLast(ListAdapter)
        .addLast(MapAdapter)
        .also { it.asConfigurable().withStandardMappings() }
        .done()
)

fun Moshi.Builder.done(): Moshi.Builder = this
    .add(MetadataKotlinJsonAdapterFactory())
    .add(Unit::class.java, UnitAdapter)  // UnitAdapter is private, and must be copied or made public
```

This is a very simple change, so I don't know if it deserves to be merged into `http4k-format-moshi`, made into a new `http4k-format-moshi-lite` module, or just posted as a cookbook recipe.  There's also the matter that several other http4k modules directly depend on kotlin-reflect, such as `http4k-contract`.  If you wanted to eliminate kotlin-reflect, you could still use `http4k-contract` to share contracts, but you would be unable to generate an openapi spec.  Thoughts?

daviddenton:
This is a really great find and definitely something which requires more investigation! :)

The claim seem to be that the metadata version of moshiX will replace kotlin-reflect - which would be amazing  - so I'm up for making that change in `http4k-format-moshi` if we believe that there are no other downsides - do we have any evidence either way?

[...]

PR #765, opened 2022-08-08
replace moshi-kotlin with moshi-metadata-reflect

oharaandrew314:
As discussed in #763.  Replaces the [moshi-kotlin](https://github.com/square/moshi/tree/master/moshi-kotlin) module (based on `kotlin-reflect`) with the [moshi-metadata-reflect](https://github.com/ZacSweers/MoshiX/tree/main/moshi-metadata-reflect) module, based on `kotlinx-metadata`.  The new reflection backend is supposed to bring various performance improvements, but it also reduces the jar size from 3 MB to 1 MB, which is incredibly useful in serverless environments.

- Tested on Android
- Http4k tests pass
- My own small pet project is successfully using this

### DIFF
--- a/http4k-format/moshi/build.gradle.kts
+++ b/http4k-format/moshi/build.gradle.kts
@@ -4,7 +4,7 @@ dependencies {
     api(project(":http4k-format-core"))
     api(project(":http4k-realtime-core"))
     api(Square.moshi)
-    implementation("dev.zacsweers.moshix:moshi-metadata-reflect:0.18.3")
+    api("dev.zacsweers.moshix:moshi-metadata-reflect:0.18.3")
     testImplementation(project(":http4k-core"))
     testImplementation(project(path = ":http4k-core", configuration ="testArtifacts"))
     testImplementation(project(path = ":http4k-format-core", configuration ="testArtifacts"))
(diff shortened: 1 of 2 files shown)


oharaandrew314:
@daviddenton if you're getting cold feet about this fairly drastic change, shall I withdraw the PR?

daviddenton:
I'd love to get this in if it was stable, but am too nervous about it to make it into Moshi proper. 

I wonder if another module is better, or the incubator or just a docs update.

Thanks so much for highlighting it though. I think we should put some effort to try and get the thing put into moshi...

oharaandrew314:
If we want to go the docs route, I can either:

- Add the instructions to [Deploying an http4k app to AWS Lambda](https://www.http4k.org/guide/tutorials/serverless_http4k_with_aws_lambda/)
- Make a new How-to Guide
- Both

daviddenton:
I think we should keep the tutorials as simple as possible, so maybe just a quick section in the how-to would be enough

Maybe... "How to avoid using reflection for JSON marshalling?" 😉

oharaandrew314:
AFAIK, `moshi-metadata-reflect` is still reflection, but I'll get a how-to written soon.

PR #776, opened 2022-09-09
Nonblocking websocket client to receive binary messages

### LINKED ISSUE #775: WebsocketClient.nonBlocking cannot receive messages in binary mode
The non-blocking websocket client does not seem to receive binary mode messages.  Send works fine.

This failing test is excerpted from `WebsocketClientTest`, so is missing some of the shared setup code.

```kotlin
val ws = websockets(
    "/bin" bind { ws: Websocket ->
        ws.onMessage {
            val content = it.body.stream.readBytes()
            ws.send(WsMessage(content.inputStream()))
            ws.close(NORMAL)
        }
    }
)

@Test
fun `non-blocking - send and receive in binary mode`() {
    val queue = LinkedBlockingQueue<() -> WsMessage?>()
    val received = generateSequence { queue.take()() }

    WebsocketClient.nonBlocking(Uri.of("ws://localhost:$port/bin")) { ws ->
        ws.onMessage { message ->
            queue.add { message }
        }
        ws.send(WsMessage("hello".byteInputStream()))
        ws.onClose {
            queue.add { null }
        }
    }

    val messages = received.take(4).toList()
    assertThat(messages, hasSize(equalTo(1)))
    assertThat(messages.first().body is StreamBody, equalTo(true))
    assertThat(messages.first().bodyString(), equalTo("hello"))
}
```

> expected: a value that has size that is equal to 1
but had size that was: 0
java.lang.AssertionError: expected: a value that has size that is equal to 1
but had size that was: 0

oharaandrew314:
Closes #775

### DIFF
--- a/http4k-client/websocket/src/main/kotlin/org/http4k/client/internal/internal.kt
+++ b/http4k-client/websocket/src/main/kotlin/org/http4k/client/internal/internal.kt
@@ -57,6 +57,10 @@ class NonBlockingClient(uri: Uri, headers: Headers, timeout: Duration, private v
 
     override fun onMessage(message: String) = socket.get().triggerMessage(WsMessage(message))
 
+    override fun onMessage(bytes: ByteBuffer) {
+        socket.get().triggerMessage(WsMessage(Body(bytes)))
+    }
+
     override fun onError(e: Exception) = socket.get().triggerError(e)
 }
 
--- a/http4k-client/websocket/src/main/kotlin/org/http4k/client/internal/internal.kt
+++ b/http4k-client/websocket/src/main/kotlin/org/http4k/client/internal/internal.kt
@@ -57,9 +57,7 @@ class NonBlockingClient(uri: Uri, headers: Headers, timeout: Duration, private v
 
     override fun onMessage(message: String) = socket.get().triggerMessage(WsMessage(message))
 
-    override fun onMessage(bytes: ByteBuffer) {
-        socket.get().triggerMessage(WsMessage(Body(bytes)))
-    }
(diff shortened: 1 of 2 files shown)


http4k-client/websocket/src/main/kotlin/org/http4k/client/internal/internal.kt
```diff
@@ -57,6 +57,8 @@ class NonBlockingClient(uri: Uri, headers: Headers, timeout: Duration, private v
 
     override fun onMessage(message: String) = socket.get().triggerMessage(WsMessage(message))
 
+    override fun onMessage(bytes: ByteBuffer) = socket.get().triggerMessage(WsMessage(Body(bytes.array().inputStream())))
```
oharaandrew314, line 60: I don't like assuming what's in a `ByteBuffer`, but the `BlockingQueueClient` does the same, and will ensure the body is ` StreamBody` as expected.

http4k-client/websocket/src/test/kotlin/org/http4k/client/WebsocketClientTest.kt
```diff
@@ -135,4 +145,24 @@ class WebsocketClientTest {
...
+                queue.add { message }
+            }
+            ws.onClose {
+                queue.add { null }
+            }
+            ws.send(WsMessage("hello".byteInputStream()))
+        }
+
+        val messages = received.take(4).toList()
+        assertThat(messages.all { it.body is StreamBody }, equalTo(true))
```
oharaandrew314, line 165: Please feel free to suggest a nicer way to do this instanceOf assertion with hamkrest.

daviddenton: not sure... :)

PR #780, opened 2022-09-19
allow netty ws config to be overridden

oharaandrew314:
I don't love how this complicates the nice simple `Netty` constructor, but I think it's valid to want to override some default settings.

### DIFF
--- a/http4k-server/netty/src/test/kotlin/org/http4k/websocket/NettyWebsocketTest.kt
+++ b/http4k-server/netty/src/test/kotlin/org/http4k/websocket/NettyWebsocketTest.kt
@@ -2,5 +2,17 @@ package org.http4k.websocket
 
 import org.http4k.client.JavaHttpClient
 import org.http4k.server.Netty
+import org.http4k.server.WebSocketServerHandler
+import org.junit.jupiter.api.Test
 
-class NettyWebsocketTest : WebsocketServerContract(::Netty, JavaHttpClient())
+class NettyWebsocketTest : WebsocketServerContract(::Netty, JavaHttpClient()) {
+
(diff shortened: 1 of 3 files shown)


http4k-server/netty/src/test/kotlin/org/http4k/websocket/NettyWebsocketTest.kt
```diff
@@ -2,5 +2,17 @@ package org.http4k.websocket
...
+import org.junit.jupiter.api.Test
 
-class NettyWebsocketTest : WebsocketServerContract(::Netty, JavaHttpClient())
+class NettyWebsocketTest : WebsocketServerContract(::Netty, JavaHttpClient()) {
+
+    @Test
+    fun `override websocket config`() {
+        Netty(8000, createWebsocketHandler = { ws ->
+            WebSocketServerHandler(ws) {
+                it.maxFramePayloadLength(30_000)
```
oharaandrew314, line 14: I know this test doesn't actually test anything, but it ensures this use-case compiles.

daviddenton:
Yeah - to be honest I'm not a massive fan of this. Part of the reason for us not providing custom config was to avoid having o(n) options on creating a server. I wonder if there is a better way to do this so that people can more easily compose custom servers?

oharaandrew314:
I don't like it either.  The composition idea is interesting though; I might try to come up with a proposal around it.  In hindsight, I suppose the `PolyHandler` concept is kind of clunky.  I believe this is especially relevant to Netty application, but we'll have to take special care to ensure whatever composite is proposed will ensure thread pools aren't duplicated for each component of the server.

PR #783, opened 2022-09-22
Add Coroutine context to Zipkin traces

alan-uk:
Allows the Zipkin client and server filters to be used in a coroutine context, so the trace ID is still propagated even if the outgoing request is on a different thread. To be used like:

runBlocking(ZipkinTraces.asContextElement()) {
 ... 
}

### DIFF
--- a/http4k-core/build.gradle.kts
+++ b/http4k-core/build.gradle.kts
@@ -5,6 +5,7 @@ dependencies {
     implementation("javax.servlet:javax.servlet-api:_")
     implementation("dev.forkhandles:result4k:_")
     implementation("dev.forkhandles:values4k:_")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4")
 
     testApi(project(":http4k-client-apache"))
     testApi(project(":http4k-client-websocket"))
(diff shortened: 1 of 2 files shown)


s4nchez:
Hi @alan-uk. Can you expand on your use case for ZipkinTraces inside coroutines?

We're happy to consider a contribution on that topic, but don't see a compelling reason to add coroutine dependency to the core of http4k.

alan-uk:
Hi Ivan,

The use case we have is where we have an incoming API which needs to call a number of other services to collect information. To help with performance we wanted to call these in parallel rather than sequentially, using the coroutine model. However this breaks the mechanism that the ZipkinTraces code uses to pass on Trace IDs. It stores them in a ThreadLocal object, so if the outgoing API is called on a different thread, the Trace ID is lost.

ThreadLocal has a method asContextElement() that can be passed to runBlocking() when setting up a coroutine context that ensures that the ThreadLocal data is shared between any threads that are used to run code in the coroutine context. However we can’t use this directly because the ThreadLocal object (THREAD_LOCAL) is marked as internal and so is inaccessible outside the http4k code.

So to resolve this, we can either:

• Add a method to ZipkinTraces to access the asContextElement() method (my PR)
• Make the THREAD_LOCAL itself accessible
• Allow a user-provided storage implementation to be used instead of THREAD_LOCAL

Regards
Alan Wilson
(Sedex Information Exchange)

s4nchez:
Hi @alan-uk, thank you for the context and the possible options that would work for your context. 

We've added a `ZipkinTracesStorage` interface which can be injected into the relevant filters (available in 4.31.0.0 which is being released now). 

Please give it a try and let us know if that solves your problem.

Cheers!

alan-uk:
That’s great news thanks, I’ll try it out.

Alan

PR #789, opened 2022-10-02
fix Sponsor Us! link

p10r:
Fixes a link which is currently broken:

image


daviddenton:
Thanks! Does this.mean you were unable to press it? 😉

PR #790, opened 2022-10-02
cap line length for .kt files to 80 in src/docs

p10r:
[Some of the docs](https://www.http4k.org/guide/howto/configure_an_oauth_server/) currently look like this:

image

One has to use horizontal scrolling to see the whole code (Firefox, MBP 2019, no zoom)

Suggestion: Applying `max_line_length = 80` for all Kotlin files in `src/docs`, so everything fits on the screen without scrolling.

Searched around for this and it seems like this was already a thing in the [slack channel](https://kotlinlang.slack.com/archives/C5AL3AKUY/p1612697628109000?thread_ts=1612662459.103400&cid=C5AL3AKUY), but seems like the PR was never opened(?).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,3 +20,5 @@ ij_kotlin_name_count_to_use_star_import = 99
 ij_kotlin_name_count_to_use_star_import_for_members = 99
 disabled_rules = chain-wrapping,colon-spacing,comma-spacing,comment-spacing,curly-spacing,filename,import-ordering,indent,indent_size,no-unused-imports,no-wildcard-imports,parameter-list-wrapping,range-spacing,string-template
 
+[src/docs/*/**.kt]
+max_line_length = 80
--- a/.editorconfig
+++ b/.editorconfig
@@ -21,4 +21,4 @@ ij_kotlin_name_count_to_use_star_import_for_members = 99
 disabled_rules = chain-wrapping,colon-spacing,comma-spacing,comment-spacing,curly-spacing,filename,import-ordering,indent,indent_size,no-unused-imports,no-wildcard-imports,parameter-list-wrapping,range-spacing,string-template
 
 [src/docs/*/**.kt]
-max_line_length = 80
+max_line_length = 95
--- a/src/docs/blog/typesafe_configuration/post/typesafe.kt
+++ b/src/docs/blog/typesafe_configuration/post/typesafe.kt
@@ -17,9 +17,12 @@ data class Host(val value: String)
 
 val env = Environment.ENV
 
-val portLens: Lens<Environment, Port> = EnvironmentKey.int().map(::Port).defaulted("PORT", Port(9000))
-val timeoutLens: Lens<Environment, Timeout?> = EnvironmentKey.duration().map(::Timeout).optional("TIMEOUT")
-val hostsLens: Lens<Environment, List<Host>> = EnvironmentKey.map(::Host).multi.required("HOSTNAMES")
+val portLens: Lens<Environment, Port> =
+    EnvironmentKey.int().map(::Port).defaulted("PORT", Port(9000))
+val timeoutLens: Lens<Environment, Timeout?> =
+    EnvironmentKey.duration().map(::Timeout).optional("TIMEOUT")
+val hostsLens: Lens<Environment, List<Host>> =
+    EnvironmentKey.map(::Host).multi.required("HOSTNAMES")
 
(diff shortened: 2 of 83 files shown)


p10r:
Anyway, wanted to check back before I go through every file and check what the IDEA formatter did.
What do you think about altering the `.editorconfig` for this?

daviddenton:
Yeah, seems like a legit change. Happy to do it if we can limit the length just in the docs folder (I think we use 120 elsewhere)

p10r:
Yeah, checked it and it's working as intended. It's also only targeting Kotlin files, nothing else.

Let me take care of it and have a look later on. 👍

.editorconfig
```diff
@@ -20,3 +20,5 @@ ij_kotlin_name_count_to_use_star_import = 99
 ij_kotlin_name_count_to_use_star_import_for_members = 99
 disabled_rules = chain-wrapping,colon-spacing,comma-spacing,comment-spacing,curly-spacing,filename,import-ordering,indent,indent_size,no-unused-imports,no-wildcard-imports,parameter-list-wrapping,range-spacing,string-template
 
+[src/docs/*/**.kt]
+max_line_length = 80
```
p10r, line 24: Most important part

src/docs/blog/typesafe_configuration/pre/secrets.kt
```diff
@@ -7,9 +7,10 @@ import java.nio.ByteBuffer
 
 val s3 = OkHttp()
 
-fun readFile(secretKey: String, bucketKey: String): ByteBuffer {
-    return s3(Request(GET, "https://mybucket.s3.amazonaws.com/$bucketKey")).body.payload
-}
+fun readFile(secretKey: String, bucketKey: String): ByteBuffer =
```
p10r, line 10: `secretKey` is not being used - not sure if intended. ([Corresponding Docs](https://www.http4k.org/blog/typesafe_configuration/#4_security)

razvn:
Please not 80 we're no longer in old Unix system, we have 4k-6k screens I hate all the white space on the web because everyone thing we are still in 1024px...

Anyway 120 is what fits on a standard HD screen with left and right panels of IntelliJ so IMHO is a minimum and a good enough compromise.

daviddenton:
@razvn this is only for the website docs content, so am happy to go with whatever fits on most people's screens. Seems to be about 95 on my screen. Definitely NOT the IDE 😄

p10r:
Hey @razvn, it's not about the general code style in the repo, it's just about the docs that are being shown when [browsing the docs](https://www.http4k.org/guide/howto/configure_an_oauth_server/). There's a lot of horizontal scrolling:
image

p10r:
Here's a better example with text above it and full screen: 
image

p10r:
We could also go for 95 @daviddenton - should also work fine. Wdyt?

daviddenton:
Yeah - 95 works better. Will be reasonable to code against whilst being readable.

Thanks for picking this up BTW @p10r  - docs are important! 😄

src/docs/blog/typesafe_configuration/post/typesafe.kt
```diff
@@ -17,9 +17,12 @@ data class Host(val value: String)
 
 val env = Environment.ENV
 
-val portLens: Lens<Environment, Port> = EnvironmentKey.int().map(::Port).defaulted("PORT", Port(9000))
-val timeoutLens: Lens<Environment, Timeout?> = EnvironmentKey.duration().map(::Timeout).optional("TIMEOUT")
-val hostsLens: Lens<Environment, List<Host>> = EnvironmentKey.map(::Host).multi.required("HOSTNAMES")
+val portLens: Lens<Environment, Port> =
+    EnvironmentKey.int().map(::Port).defaulted("PORT", Port(9000))
```
p10r, line 21: We could also go for 
```kotlin
val portLens: Lens<Environment, Port> = EnvironmentKey.int()
    .map(::Port)
    .defaulted("PORT", Port(9000))
```
But, if I remember correctly, the rest of the code base is not splitting lines after each method call, right?

p10r:
> Thanks for picking this up BTW @p10r - docs are important! 😄

This is also for my own reading experience, so happy to jump in there. 🌚

p10r:
I also went through the markdown files. Looks good so far - I'll create a follow-up if I ever find another occurence.

daviddenton:
thanks again :)

p10r:
Thanks for building such a library ❤️

PR #793, opened 2022-10-04
Add UserCredentialsOAuthSecurity and OpenApi3 renderer

oharaandrew314:
The OAuth grant is called `Resource Owner Password Credentials`, but the existing `ClientFilter` has already set precedent by calling it `UserCredentials`.  I've decided to defer to it.


oharaandrew314:
@daviddenton Do you want this included?

PR #798, opened 2022-10-14
Allow for keeping temporary multiparts around.

jippeholwerda:
When uploading a file through a multipart form, we want to keep the temp file around for async virus scanning. However, in the following piece of code, the file is deleted upon JVM exit:
```kotlin
    private fun writeToDisk(part: StreamingPart, bytes: ByteArray, length: Int) =
        File.createTempFile(part.fileName ?: UUID.randomUUID().toString()
        + "-", ".tmp", temporaryFileDirectory).apply {
            deleteOnExit()
            FileOutputStream(this).apply {
                write(bytes, 0, length)
                use { part.inputStream.copyTo(it, writeToDiskThreshold) }
            }
        }
```
with the `deleteOnExit()` call.

We want to keep the file around even when the JVM exits (we're running on k8s so pods can always be killed).

This PR uses a flag that is passed from MultipartFormBody.from to the MultipartFormParser indicating to delete (or not) the temp file, delegating control over when the temp file will be deleted to the caller.

### DIFF
--- a/http4k-multipart/src/main/kotlin/org/http4k/core/MultipartFormBody.kt
+++ b/http4k-multipart/src/main/kotlin/org/http4k/core/MultipartFormBody.kt
@@ -82,7 +82,7 @@ data class MultipartFormBody private constructor(internal val formParts: List<Mu
     companion object {
         const val DEFAULT_DISK_THRESHOLD = 1000 * 1024
 
-        fun from(httpMessage: HttpMessage, diskThreshold: Int = DEFAULT_DISK_THRESHOLD, diskDir: File = Files.createTempDirectory("http4k-mp").toFile().apply { deleteOnExit() }): MultipartFormBody {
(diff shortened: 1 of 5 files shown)


daviddenton:
We can support something like this, but I think we can do something a little cleverer - passing a Boolean is a bit yuck. Maybe a factory to provide the file (which would then default to one creating Temp files) would be better?

At the very least, it should be a non-breaking change by putting the option at the end of the method Sig.

jippeholwerda:
Agreed. I'll see if I can come up with something better.

jippeholwerda:
Anything I can do to move this forward?

daviddenton:
We just need to have a closer look at it. It looks all ok from first glance though. Shouldn't be too long - just a matter of getting a few minutes

jippeholwerda:
👍🏻

s4nchez:
Thank you, @jippeholwerda. 

That'll be available in 4.32.4.0 😄

daviddenton:
FYI We've actually made a couple of changes to the API and have released these as 4.33.0.0. There is also a CVE fix for handlebars in there :)

jippeholwerda:
Cool! We needed that as well :-)

PR #802, opened 2022-10-19
Supply scopes for oauth requests without oauthprovider

p10r:
Adds support for providing the [OAuth Scope](https://oauth.net/2/scope/) when using the `OAuthClientFilterExtensions`.


p10r:
That's just a first draft - there are still some rough edges.

http4k-security/oauth/src/main/kotlin/org/http4k/security/oauth/client/clientFilterExtensions.kt
```diff
@@ -99,14 +104,27 @@ fun ClientFilters.OAuthUserCredentials(clientCredentials: Credentials, userCrede
...
-                    .with(
+                if (scopes.isNotEmpty())
+                    requestForm of WebForm().with(
+                        grantType of "client_credentials",
+                        clientId of clientCredentials.user,
+                        clientSecret of clientCredentials.password,
+                        scope of scopes.joinToString(separator = " ")
+                    )
+                else
+                    requestForm of WebForm().with(
```
p10r, line 127: Would've loved to no have that duplication, but couldn't come up with another way

http4k-security/oauth/src/main/kotlin/org/http4k/security/oauth/client/clientFilterExtensions.kt
```diff
@@ -67,16 +70,18 @@ fun ClientFilters.RefreshingOAuthToken(
 fun ClientFilters.RefreshingOAuthToken(
     config: OAuthProviderConfig,
     backend: HttpHandler,
-    oAuthFlowFilter: Filter = ClientFilters.OAuthClientCredentials(config.credentials),
+    oAuthFlowFilter: Filter = ClientFilters.OAuthClientCredentials(config.credentials, emptyList()),
     gracePeriod: Duration = Duration.ofSeconds(10),
-    clock: Clock = Clock.systemUTC()
+    clock: Clock = Clock.systemUTC(),
+    scopes: List<String> = emptyList(),
```
p10r, line 76: I guess it generally makes also sense to have an option to supply the scope for `ClientFilters.OAuthCredentials()`.
The problem is that the scope has to be provided again at for `ClientFilters.RefreshingOAuthToken`.
We could move `scopes` up in the parameter list, but this would introduce a breaking change.

http4k-security/oauth/src/main/kotlin/org/http4k/security/oauth/client/clientFilterExtensions.kt
```diff
@@ -42,21 +41,25 @@ fun ClientFilters.RefreshingOAuthToken(
     oauthCredentials: Credentials,
     tokenUri: Uri,
     backend: HttpHandler,
-    oAuthFlowFilter: Filter = ClientFilters.OAuthClientCredentials(oauthCredentials),
+    oAuthFlowFilter: Filter = ClientFilters.OAuthClientCredentials(oauthCredentials, emptyList()), //TODO feels weird
```
p10r, line 44: see other comment

http4k-security/oauth/src/main/kotlin/org/http4k/security/oauth/client/clientFilterExtensions.kt
daviddenton: If scope is an optional lens you can maybe do scope.takeif{it.isNotempty()}.jointostring...

http4k-security/oauth/src/main/kotlin/org/http4k/security/oauth/client/clientFilterExtensions.kt
p10r: For now I went with highlighting that no scopes are passed into the oAuthFlowFilter - this way it's not breaking.
What do you think?

p10r:
Please have a look if you find the time @daviddenton . Will create an example for the blog once we settled on a solution. 👍

http4k-security/oauth/src/test/kotlin/org/http4k/security/oauth/client/OAuthUserCredentialsTest.kt
```diff
@@ -18,8 +18,31 @@ class OAuthUserCredentialsTest {
...
+                    "&client_secret=world" +
+                    "&username=user" +
+                    "&password=password" +
+                    "&scope=someactivity.write+someactivity.read"
+            )
+        )
+    }
+
+    @Test
+    fun `omits scope if not provided`() {
```
p10r, line 40: Those are kind of duplicates, but since every filter uses a different set of parameters I rather went with a version for everyone of them

PR #803, opened 2022-10-20
Recipe for "moshi-lite" via kotlinx-metadata-reflect

### LINKED ISSUE #763: Moshi - Replace Kotlin-Reflect with kotlinx-metadata-jvm

oharaandrew314:
Alternate solution to close #763


build.gradle.kts
```diff
@@ -233,6 +233,8 @@ dependencies {
             testImplementation(project(path = it.name, configuration = "testArtifacts"))
         }
 
+    implementation("dev.zacsweers.moshix:moshi-metadata-reflect:0.19.0")
```
s4nchez, line 236: Can't this live in `http4k-format/mochi/build.gradle.kts` rather than the root gradle config?

daviddenton: this is a good point - although it is consistent with our other examples. All of the dependencies for the docs live in the root module at the moment. :)

s4nchez: Ah! That's the dependency for example! I thought it was the actual one for the mochi module. Please ignore me...

PR #805, opened 2022-10-20
Jetty WebSocket blocking and non-blocking clients

FredNordin:
Adds the ability to create blocking (WsClient) and non-blocking (Websocket) websocket clients using the Jetty websocket client library, instead of the reference java implementation. Useful when the Jetty HTTP client is already used in an application and one wants to use the improved functionality and performance it offers.

This PR also introduces a couple of test contracts for blocking vs non-blocking clients and the existing WS client tests have been migrated to use them.


oharaandrew314:
This is good stuff.  I happen to have a prototype app that uses both WebSockets AND raw TCP.  Even though it's highly doubtful we'll provision a JVM app on actual devices, it will still be good to use Netty for both connection types.

PR #816, opened 2022-11-15
Add Json utility methods to define mapping from json to objects

FredNordin:
This PR adds the ability to manually define mapping from JSON to objects using a set of new utility methods in `Json` interface.

This is useful when we don't know what implementation we are using and cannot use auto-marshalling functionality from e.g. Jackson or Moshi.

This is in the similar vein as the current utility methods to create JSON.

Naming of methods etc are all open for discussion - I simply went with what I thought made the most sense - but there are probably better names.

I am raising this as a draft PR, so that I can gather feedback on the implementation and if it is actually a good idea or not.

NOTE: I also had to change the signature of `Json<NODE>` to be `Json<NODE : Any>` to make it work with lenses, which do require the generic type to be `IN : Any`. That is why there are so many files changed, even though all the core changes are in `Json` and the associated tests.

### DIFF
--- a/http4k-format/core/src/main/kotlin/org/http4k/format/Json.kt
+++ b/http4k-format/core/src/main/kotlin/org/http4k/format/Json.kt
@@ -7,9 +7,14 @@ import org.http4k.lens.BiDiBodyLensSpec
 import org.http4k.lens.BiDiLensSpec
 import org.http4k.lens.ContentNegotiation
 import org.http4k.lens.ContentNegotiation.Companion.None
+import org.http4k.lens.LensGet
+import org.http4k.lens.LensSet
 import org.http4k.lens.Meta
 import org.http4k.lens.MultipartFormField
+import org.http4k.lens.ParamMeta
+import org.http4k.lens.ParamMeta.ArrayParam
 import org.http4k.lens.ParamMeta.ObjectParam
+import org.http4k.lens.ParamMeta.StringParam
 import org.http4k.lens.httpBodyRoot
 import org.http4k.lens.string
 import org.http4k.websocket.WsMessage
@@ -19,7 +24,7 @@ import java.math.BigInteger
 /**
  * This is the contract for all JSON implementations
  */
-interface Json<NODE> {
+interface Json<NODE : Any> {
     // Contract methods to be implemented
     fun NODE.asPrettyJsonString(): String
     fun NODE.asCompactJsonString(): String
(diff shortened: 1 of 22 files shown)


http4k-format/core/src/main/kotlin/org/http4k/format/Json.kt
```diff
@@ -88,8 +93,51 @@ interface Json<NODE> {
...
+    fun <T : Any> asA(mapper: (NODE) -> T): (NODE) -> T = { mapper(it.requireObject()) }
+    fun <T : Any> asArray(mapper: (NODE) -> T): (NODE) -> List<T> = { elements(it.requireArray()).map(asA(mapper)) }
+    fun field(): BiDiLensSpec<NODE, String> = fieldLens(StringParam) { name, node -> textValueOf(node, name) }
+    fun <T : Any> obj(mapper: (NODE) -> T): BiDiLensSpec<NODE, T> = fieldLens(ObjectParam) { _, node -> mapper(node) }
+    fun <T : Any> array(itemType: ParamMeta, mapper: (NODE) -> T): BiDiLensSpec<NODE, List<T>> =
+        fieldLens(ArrayParam(itemType)) { _, node -> elements(node).map(mapper) }
+    // ---
+
+    private fun <T : Any> fieldLens(paramMeta: ParamMeta, mapper: (String, NODE) -> T?): BiDiLensSpec<NODE, T> =
+        BiDiLensSpec("field", paramMeta,
```
s4nchez, line 107: @FredNordin do you need to introduce Lens at the field level to start with?

Unless the failure messages are better than what you usually get in the Body lens, I'd suggest adding this in a future iteration. Otherwise, I'm afraid we're adding something that will make debugging even harder than at the moment. /cc @daviddenton

FredNordin: Exactly the kind of feedback I was hoping for - thanks @s4nchez !
I can change it around to not use lenses and see what it looks like. I totally get your comment to try to keep it as "json" pure as possible and not involve lenses here, but maybe add them elsewhere, e.g. in the `org.http4k.lens` package.

FredNordin: If lenses are even needed for this, obviously...

s4nchez: > maybe add them elsewhere, e.g. in the org.http4k.lens package
If that's doable, I'd try that.

Also, I suggest including tests for failure scenarios, so we can get a feel for the kind of user feedback people could get in those cases.

FredNordin: Comprehensive set of tests will be added as soon as we're agreed on the approach and implementation.

FredNordin:
Closing this PR as I've run in to some issues that make me question the feasibility of implementing this in the way I want to (not a fault with http4k - just the way JSON is). I might resurrect this in the future if I have a lightbulb moment.

s4nchez:
No problem, @FredNordin. Thank you for giving it a try :)

PR #830, opened 2022-12-15
fix: avoid altering path params

tkint:
Hello !

When using contracts, there is a bug when parsing parameters and passing them to the handler.

For example, if you have a route like this `"somePrefix" / Path.of("value")`, when matching `/somePrefix/somePrefixInValue`, the prefix will be removed from the value.

This PR is fixing this bug by trimming only the first occurrence when matching the uri.

### DIFF
--- a/http4k-contract/src/main/kotlin/org/http4k/contract/PathSegments.kt
+++ b/http4k-contract/src/main/kotlin/org/http4k/contract/PathSegments.kt
@@ -40,4 +40,5 @@ object Root : PathSegments() {
 internal fun Request.isIn(contractRoot: PathSegments) = pathSegments().startsWith(contractRoot)
 
 internal fun Request.pathSegments() = PathSegments(uri.path)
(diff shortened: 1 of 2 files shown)


s4nchez:
@tkint that behaviour is by design. 

For the majority of cases, APIs have endpoints such as `/users/12345`, and we want to be able to extract `12345` into an object (e.g. an `UserId`). If we captured the whole URL, people would have to parse it again inside the handler.

If you need the full Uri, you can still get it from `Request.uri`. 

I hope this makes sense. Please let us know if you have a specific use case that can't be covered by the above. Otherwise, I'm closing this as it's not a bug.

tkint:
I don't understand the problem of using a `replaceFirst` instead of a `replace`. It still remove `/users` from the url, it just won't do it multiple times.

The issue is that `/users/users_12345` would give me `_12345` instead of `users_12345`.

s4nchez:
Ha - now I understand the problem. I thought you wanted to include the prefix in the URL parameter, not the one repeated in the value. Sorry about that.

And yes, that seems like a bug, so I'll merge this. Thank you for spotting and fixing it!

PR #835, opened 2022-12-22
New module providing filter that integrates with Failsafe for fault tolerance

FredNordin:
Arguably easier to use than Resilience4J and allows for multiple policies to be configured using only one Filter. It also has zero dependencies and only requires Java 8 (Resilience4J seems to have upped that to Java 17+).

I've also added a page to the reference documentation to describe the usage.


FredNordin:
The build failed because of a test in http4k-client-websocket timing out, which is not related to the changes in this PR. I cannot re-trigger the build though, so I'll leave that to one of you admin level peeps to sort out 😄

daviddenton:
Thanks! Good job :)

PR #838, opened 2023-01-06
Allow access-control-max-age header to be set from cors policy

moddular:
Allow caching of cors preflight checks as described [here](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) by letting CorsPolicy set an access-control-max-age header.


daviddenton:
Thanks!

daviddenton:
It's releasing now :)

moddular:
Thanks!

PR #841, opened 2023-01-11
Added methods for copying `ContractRoute` with new fields.

MarcusDunn:
Not sure if this exposes too much internal API - but I have what I think is a valid use case for this.

Each one of the handlers for my app is injected into a main class that creates the `contract`. I have a global security policy that I used - but I'd like the rendered schema to have this global security policy applied to each route for better client auto generation. So I need a way to "edit" (immutably of course) a `ContractRoute` such that I can add the Security to the individual routes `meta` before I add them to the final contract.


daviddenton:
I think we can go one better. If you set a security on the global contract, the child contract routes automatically inherit the global policy unless overridden in the individual route - note you can also set the Security to null (although I'm adding a `Security.None` option in the latest release so that will look nicer. 😉

MarcusDunn:
you're absolutely right - added an extra securityRenderer and all is well.

PR #843, opened 2023-01-11
Map properties should be nullable by default fixes #842

### LINKED ISSUE #842: Map OpenAPI implementation adds all properties as required
The `AutoJsonToJsonSchema` implementation sets the properties on a `map` to required, when I think they should probably be nullable. I'll submit a PR to this effect.

BBB:
:wave: Hi Folks, not sure if this is a breaking change or not, but I think this should be the default :)


daviddenton:
This is a good change and an oversight in the original implementation - maps are by definition made up of optional fields. We  should bump the "b" digit (in a.b.c.d) to communicate it to the API users.

Thanks!

PR #844, opened 2023-01-12
Add SwaggerUi oauth2-redirect and OIDC Contract Security

oharaandrew314:
- The http4k-contract `swaggerUI` helper now serves oauth2-redirect.html, required for oauth redirects
- Added http4k-contract `OpenIdConnectSecurity`

The intention here was to support Google Sign In.  Adding OAuth2 redirect  support to the `swaggerUi` helper was mandatory, and this allowed me to use the standard OAuth flows successfully.  However, I needed an `id_token` from Google, which, according to Swagger UI, can only be retrieved using OIDC.  Unfortunately, Google's OIDC `urn:ietf:params:oauth:grant-type:jwt-bearer` grant type isn't supported by Swagger UI, but at least `authorization_code` works through OIDC.  I didn't get what I wanted, but these new features can still stand on their own.


http4k-contract/src/main/resources/org/http4k/contract/ui/public/oauth2-redirect.html
```diff
@@ -0,0 +1,81 @@
+<!DOCTYPE html>
+<html lang="en-US">
+<head>
+  <title>%%PAGE_TITLE%%</title>
+</head>
+<body>
+<script>
```
oharaandrew314, line 7: I wish I could just load this script from the CDN (unsure about license implications and just maintainability in general), but even the CDN has the entire JS content embedded in the HTML.  I suppose the only other alternative would be to have the HttpHandler server the HTML directly from the CDN.

daviddenton:
Apart from the license question, this is all good I think 😄

oharaandrew314:
So, the file I copied is in the swagger-ui repo [here](https://github.com/swagger-api/swagger-ui/blob/7e65155a7bbc0bf840b4ab56a35ee794e89dfbc9/dist/oauth2-redirect.html) which is Apache 2.0 licensed.  Based on my understanding of the license, we should be free to modify and distribute their code.  Alternatively, I can just have the  `HttpHandler` forward the request to the CDN.  This would break if the server is in a private subnet, with no public internet access.  What do you think?

daviddenton:
For the moment, let's leave it as needing public internet access - that's going to be the 99% case and will leave us with less code to maintain. 😄

oharaandrew314:
To be clear; that's public access on the server, not the browser.  This tends to be frowned upon in environments that need to conform to compliance standards.

daviddenton:
Ah - gotcha. In that case, leaving the static file in there is probably best :)

oharaandrew314:
Whoops!  Bad merge.  All good to go now.

daviddenton:
👍

PR #848, opened 2023-01-24
HTTP Message Parser fixes

oharaandrew314:
I've discovered 2 issues preventing me from accurately parsing messages as they are received.

**Header values with optional leading whitespace**
The MDN specification implies that leading whitespace for header values is optional.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

**Body line-breaks not being preserved**


daviddenton:
Seems legit. Merge it! :)

PR #850, opened 2023-01-25
parsed http messages to preserve newlines

oharaandrew314:


oharaandrew314:
On my machine, I can't reproduce the  `uri is passed to the upgrade request` failure, but `onError is called on connection error` fails even on master.  Is this test flakey at all?  Or is it just a platform-specific failure that I need to account for?

PR #855, opened 2023-02-04
Standardize AutoMarshalling roundtrip of arbitrary collections

oharaandrew314:
This might be enough to use `KotlinxSerialization` with an `OpenApi3Renderer`. 

I wasn't bothered enough to make it work for klaxon 🤷


daviddenton:
Fine about Klaxon - it's a bit of an outlier anyway :)

PR #856, opened 2023-02-05
fix error parsing colons in header values

oharaandrew314:


oharaandrew314:
Already fixed by David on main

PR #871, opened 2023-03-19
Add functions to customize standard moshi and jackson

oharaandrew314:
Functions like this are useful when you want to register `Values4K` mappings without having to duplicate code that's already in the prebuilt `Jackson`, `Moshi`, etc.  For example:

```kotlin
val customMoshi = Moshi.custom {
  value(MyValueClass)
}

val myLens = customMoshi.autoBody<DtoWithMyValueClass>().toLens()
```


http4k-format/moshi/src/main/kotlin/org/http4k/format/ConfigurableMoshi.kt
```diff
@@ -121,7 +121,7 @@ open class ConfigurableMoshi(
 }
 
 fun Moshi.Builder.asConfigurable(
-    kotlinFactory: JsonAdapter.Factory = KotlinJsonAdapterFactory()
+    kotlinFactory: JsonAdapter.Factory? = null
```
daviddenton, line 124: Shall we default this to `KotlinJsonAdapterFactory()`?

daviddenton: ie. is there a reason NOT to?

daviddenton:
Looks good apart from the one question, and then will merge it 😄

http4k-format/moshi/src/main/kotlin/org/http4k/format/ConfigurableMoshi.kt
oharaandrew314: As much as I don't like defaulting arguments to null, and then replacing them with a real default, I have two reasons.

The main reason was to eliminate the repetition of defining the default both here and in `Moshi.standardConfig`.

The lesser reason is because I have thus far been able to eliminate `moshi-kotlin` as a dependency when replacing `KotlinJsonAdapterFactory` with an alternative factory.  I suspect that having this as a default argument is what's preventing it, but so far, this suspicion is unfounded.

I'll defer to you.  Let me know if you want this undone.

daviddenton: Let's keep it as a non-nullable for now, and if you manage to nail down that it is the reason then we can maybe adjust in future?

Will merge it and get it out there in the meantime.

PR #877, opened 2023-03-30
Java websocket server

oharaandrew314:


http4k-server/websocket/src/main/kotlin/org/http4k/server/websocket/JavaWebSocket.kt
```diff
@@ -0,0 +1,135 @@
...
+        val upgradeRequest = Request(Method.GET, handshake.resourceDescriptor)
+            .headers(headers)
+            .let { if (handshake.content != null) it.body(MemoryBody(handshake.content)) else it }
+            .source(RequestSource(conn.remoteSocketAddress.hostString, conn.remoteSocketAddress.port))
+
+        val wsAdapter = object : PushPullAdaptingWebSocket(upgradeRequest) {
+            override fun send(message: WsMessage) {
+                when (message.body) {
+                    is StreamBody -> conn.send(message.body.payload)
+                    else -> conn.send(message.bodyString())  // furthering the generalization that a MemoryBody is ALWAYS to use text mode
```
oharaandrew314, line 92: This has been an annoyance for me in the past.  I use a binary message format for some of my applications, and that means I must always use a `StreamBody`, or else it gets sent in text mode and corrupted.  Is this intentional?

daviddenton: TBH, I don't recall. 😂 ! If it was me, it probably seemed like the right thing to do at the time.

oharaandrew314: Well, I don't want to break the convention here.  But might be worth looking into later.

daviddenton:
Is this all ready @oharaandrew314 ? 😄

oharaandrew314:
Yes, I think it's ready to go!

PR #878, opened 2023-03-30
attempt to deflake WebsocketClientNonBlockingTest

oharaandrew314:
I can't reproduce the error on my machine, but I think this tweak might help


daviddenton:
Worth a try! thanks

PR #881, opened 2023-04-02
Redoc

### LINKED ISSUE #880: SwaggerUiWebjar relative path issue with a nested route
For posterity; from [a slack discussion](https://kotlinlang.slack.com/archives/C5AL3AKUY/p1680241840177909)

> Hi, I’ve a question regarding swagger-ui and routes.
I’ve a code similar to following:
        routes(
            "/swagger-ui" bind swaggerUiWebJar {
                url = "spec"
                pageTitle = "Cat Shelter"
                displayOperationId = true
                requestSnippetsEnabled = true
            }
        )
when I call my service in a browser with http://localhost:8000/swagger-ui I end up in an empty page. I can see swagger-ui by adding a trailing / .
How can I easily change this behaviour back, so that I can call it again without trailing slash?

This issue is due to a quirk with how browsers handle relative script paths.  You can workaround it by:
- serving the UI on the root path
- adding a trailing `/` to the browser's path

This issue should be resolved in the upcoming redoc release.

oharaandrew314:
Add lite and webjar redoc.  Also fixes #880 since the same issue also applies to redoc.


http4k-contract-ui/redoc/src/main/kotlin/org/http4k/contract/ui/redoc/redocWebjar.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.http4k.core.then
+import org.http4k.core.with
+import org.http4k.lens.Header
+import org.http4k.routing.ResourceLoader
+import org.http4k.routing.bind
+import org.http4k.routing.routes
+import org.http4k.routing.static
+
+fun redocWebjar(configFn: RedocConfig.() -> Unit = {}) = routes(
+    "" bind GET to { Response(FOUND).with(Header.LOCATION of it.uri.appendToPath("index.html")) },
```
oharaandrew314, line 18: Even though `static` will implicitly load `index.html` for a bare path, this causes issues when the HTML is attempting to load relative scripts.  Same applied to Swagger UI.

http4k-contract-ui/swagger/src/main/kotlin/org/http4k/contract/ui/swagger/swaggerUiWebjar.kt
```diff
@@ -2,18 +2,28 @@ package org.http4k.contract.ui.swagger
...
-    .also(configFn)
-    .toFilter()
-    .then(routes(
-        static(ResourceLoader.Classpath("org/http4k/contract/ui/public-config/")),
-        static(ResourceLoader.Classpath( "/META-INF/resources/webjars/swagger-ui/4.18.1"))
-    ))
+fun swaggerUiWebjar(configFn: SwaggerUiConfig.() -> Unit = {}) = routes(
+    "" bind Method.GET to { Response(Status.FOUND).with(Header.LOCATION of it.uri.appendToPath("index.html")) },
+
+    SwaggerUiConfig()
```
oharaandrew314, line 23: The `SwaggerUiConfig` filter only needs to apply to the config Javascript.  This should be a performance improvement.

http4k-contract-ui/redoc/src/main/kotlin/org/http4k/contract/ui/redoc/redocWebjar.kt
daviddenton: Nice fix 🙃

daviddenton:
Thanks again @oharaandrew314 🙃

PR #893, opened 2023-05-02
New format module based on kondor-json

FredNordin:
***PR currently in draft, because I first wanted to see if this module is something that you would be interested in accepting in to the project.***

This PR creates a new format module integrating https://github.com/uberto/kondor-json as the backing implementation.
As mentioned in the kondor-json readme, KondorJson is *"A functional Kotlin library to serialize/deserialize Json fast and safely without using reflection, annotations or code generation."*
I've recently used it in a commercial project in a http4k service and I can say it really is rather nice :)

To make it work as an automarshalling json implementation I had to introduce a very simple registry where one can register the different JsonConverter classes one would like to use. It is deliberately kept simple with only the required functionality to be able to look up a converter for the target class when creating auto lenses. I've also made sure that the number of lookups done at runtime are kept to a minimum, in most cases only once, to keep the performance as close to or the same as using "vanilla" kondor-json.

The current version used is 1.8.0 as later patch versions require java > 8. This can be updated to latest version for http4k v5 :)

Please have a look and let me know if this would of interest. If so, I will update the docs and such.


uberto (commented):
Hi, I'm Uberto, the author of Kondor-Json. Dave asked me to help with the review. ;)

First thanks for the effort! we are using Http4k in my project but not with http4k-format so I'm not familiar on what is exactly needed. I am writing this without having tested on my local machine, I plan to do it tomorrow.
On the tech side, the code is generally good and clear, but I have 2 main concerns:
1) it seems to introduce a breaking change (.customise().done()) to all the other json modules, this seems wrong to me. It's not clear why this is needed.
2) it seems to copy-paste a lot of code from Kondor-Json itself. I'm pretty sure this is avoidable, in case I'm happy to add some configuration in Kondor to make them redundant.

uberto:
Thanks to you! I'm happy other people are using it.

Ok, regarding AutoMarshallingJsonContract, I'm still not sure how it works, but it would seems easier to add a static register, like an object, and register all the converters there. Then pass it to customise(). I suppose you cannot do it but it's not clear why to me.
I will look better at the code with calm tomorrow. 
If the change won't impact the users I don't have a big objection, but still it would be better if we don't have to change the rest of tests. :)

regarding the duplication issue:  a possibility  would be to release a kondor 1.8.5 with the compactJson mode and Java8, and then a kondor 1.9.0 with Java 17. We can also merge the module and doing this stuff later anyway. Bad of me to upgrade Java in a minor release.

I also think that there is some other unnecessary code, meaning that some functionality you wrote should be available in kondor (but maybe they are not public?) but I would also fix that later rather than blocking the PR.

daviddenton:
It's worth pointing out that we are still targeting Java8 as a baseline for developing against http4k. v5 will build with Java 20, but we won't require it to run (currently) against anything but Loom, so we if Kondor requires 17 we would need to highlight that.

FredNordin:
A static register won't work because we need to be able to support people creating different ConfigurableKondorJson instances with different configuration, so each one has to have it's own register. For example, there might be one instance for a client adapter and another one for the server adapter.

I would really like to see a version 1.8.5 that supports the "flags" to the `render()` method you mentioned, as well as making it work on Java8. That would definitely be my preferred solution :+1:

FredNordin:
@uberto  It could be that I'm completely misunderstanding your point about a static register though, so some more information about your idea would be good  :)

FredNordin:
Again, as mentioned before, this whole thing might be a terrible idea, hence the PR being in draft.

uberto:
Ok, to be completely honest, I'm still unsure why we need the register at
all. I was thinking to do something lobjeike that to avoid declaring the
converters in the fields, so writing:

JAny<Whatever>{
 val user by obj(Whatever::user)
...

instead of

JAny<Whatever>{
 val user by obj(Whatever::user, JUser)
...

but this doesn't seem the case here... unless I'm missing something. Also I
gave up because when I asked the rest of the team, nobody liked it. ;)

Anyway the register I was thinking about was static and still configurable,
like:

object RegisterForClientA = build( User to JUser, ...)

object RegisterForClientB = build( User to JUserB, ...)


About version 1.8.5 with compact render and Java8, it seems reasonable to
me, so I can release it later this week.

FredNordin:
I'll have a think and play around with the concept of having a register externally and passing that to the `KondorJson` instance, instead of the current impl. It's an interesting idea that I'd like to explore further.

Additionally, another feature of kondor-json I like is that all converters know how to generate its json-schema. This sounds like something we could potentially use for the schema rendering in the `contracts` module, instead of the current best-effort or reflection based approaches. I thought I needed to make KondorJson an automarshalling json thingy for that, but might not need to. I'll explore that further too.

uberto:
We also use the schema in the OpenApi contracts, but again we have built a
few custom things.
Let me know what you think, I'm happy to collaborate and add features to
the library to support the use case, so far as they match the general
philosophy of Kondor.

thanks a lot for the discussion!
Anyway I will look at the PR code in detail in the next few days.

Uberto

uberto:
@FredNordin I have released a version of Kondor-Json 1.8.5 with Java8 and a method to render in compact Json

FredNordin:
@uberto I still get the same gradle error even when using version 1.8.5:
```
Execution failed for task ':http4k-format-kondor-json:compileKotlin'.
> Could not resolve all files for configuration ':http4k-format-kondor-json:compileClasspath'.
   > Could not resolve com.ubertob.kondor:kondor-core:{require 1.8.5; reject _}.
     Required by:
         project :http4k-format-kondor-json
      > No matching variant of com.ubertob.kondor:kondor-core:1.8.5 was found. The consumer was configured to find a library for use during compile-time, compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm' but:
          - Variant 'apiElements' capability com.ubertob.kondor:kondor-core:1.8.5 declares a library for use during compile-time, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm':
              - Incompatible because this component declares a component, compatible with Java 19 and the consumer needed a component, compatible with Java 8
          - Variant 'javadocElements' capability com.ubertob.kondor:kondor-core:1.8.5 declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
                  - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'jvm')
          - Variant 'runtimeElements' capability com.ubertob.kondor:kondor-core:1.8.5 declares a library for use during runtime, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm':
              - Incompatible because this component declares a component, compatible with Java 19 and the consumer needed a component, compatible with Java 8
          - Variant 'sourcesElements' capability com.ubertob.kondor:kondor-core:1.8.5 declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
                  - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'jvm')
```

uberto:
Incompatible because this component declares a component, compatible
with Java 19 and the consumer needed a component, compatible with Java
8


this is strange, since I just rolled back the gradle file as previous
version. Also... Java 19??? where does it come from? ;)

uberto:
@FredNordin 
It's something related to new gradle modules. I don't understand how to fix it, so for the moment I disabled in the 1.8.x branch. Finger crossed but version 1.8.7 should work.
I'm also happy to work a bit on your branch to improve kondor integration. Either before or after the merge in http4k-format.

FredNordin:
@uberto Thanks for trying to sort it out. I cannot work it out either - first I thought it was because you are building it on Java 19 in the CI env, but the GH workflow file states 17, so that's not it.
I'd be happy to collaborate on this in the branch before the merge. Just push any changes you feel are good, happy to discuss anything else :+1:

FredNordin:
@uberto Version 1.8.7 now works :partying_face: One minor thing that would be good to add to the new `compact` method, would be a parameter to control whether to render null nodes or not, like one can with the `pretty` method. Something like this:

```
fun JsonNode.compact(builder: StringBuilder = StringBuilder(), explicitNull: Boolean = false): StringBuilder = ...
```
Note that I also made `builder` defaulted - most of the time one would want a new builder anyway.

uberto:
I've tried to create a new PR against your branch without success.

I've put my changes here:
https://github.com/uberto/http4k/tree/kondor-json-format

not much, just updating kondor-json and some tidy up.

I'm still quite confusing by how bidimapping etc is supposed to work. 
For example the http4k-format-argo module is much smaller, but maybe it has less capabilities?
Anyway for me this is good for merge in http4k.

daviddenton:
BiDiMapping is there when you want to provide a auto-conversion from a JSON primitive to another type and back again - say a Uri or an Instant or a Value. Argo is smaller because it doesn't have that mapping style. You register converters (BiDiMappings) into the Format instance.

There are currently these types of Format instance:
- Json <-- this has an underlying Json node model
- AutoMarshalling <-- this automatically converts from JSON strings -> types. 
- AutomarshallingJson <-- both of the above.

Which one are we targetting?

uberto:
The patch is about AutoMarshallingJson. As for myself I'm happy with the patch as it is now applied. I've done some tidy up with the new kondor-json. That could be applied before or after.

"BiDiMapping is there when you want to provide a auto-conversion from a JSON primitive to another type and back again - say a Uri or an Instant or a Value." -> if you're using Kondor-Json you will probably already have the conversions to all your types and back (for example for values rendered as string you can create instances of JStringRepresentable, same for number etc.), so I'm not sure how the best approach would look like, you probably don't want to define same thing twice.

oharaandrew314:
As a fly on the wall, I'm intrigued by this.  I wonder if the custom converters could replace versioned DTOs to a degree 🤔 .

uberto:
Well, I hope so since I wrote Kondor with that goal. ;)
But it's ot here. We can discuss on Twitter or slack or kondor-json GH page.

FredNordin:
Not been able to look at this over the last few days, but here is my planned approach for this (shout if you disagree):

1. To able to use an `auto` lens, all we need is a simple register that can look up a JConverter by class - that's all. The current implementation to allows a user to look up any registered converter, along with the ability to register and resolve all the standard BiDiMapping definitions, is unnecessary. Instead there could be a number of pre-defined converters for those, e.g. `JStatus`, `JUri`, `JZonedDateTime` etc, that simply use the existing BiDiMapping underneath.
I know this would look slightly different to the other format modules, but I think it's acceptable to keep things more "kondor-json-ish" in this module.
2. I will keep the current render code in this module, because this module also need to be able to output null fields when rendering the compact json, to pass the tests. We can replace this in the future when the rendering functionality in kondor-json has been moved to the JConverters, as mentioned previously by Uberto.

This will hopefully keep the PR focussed on the bare minimum of having a AutoMarshallingJson format using kondor-json, that is as easy as possible to create and configure, which passes the existing tests.
Anything else, like using it in the contract module for rendering the openapi json schema, can be handled in the future.

uberto:
I totally second this

FredNordin:
@uberto I have made the changes I mentioned, as well as incorporating some of the changes you made in your branch.

uberto (approved):

http4k-format/kondor-json/src/main/kotlin/org/http4k/format/KondorJson.kt
```diff
@@ -0,0 +1,231 @@
...
+        requireNotNull(converters.find { it.canConvert(target) } as? ConverterWrapper<T, *>) {
+            "JsonConverter for '$target' has not been registered"
+        }
+
+    private fun converterFor(input: Any): ConverterWrapper<*, *> =
+        requireNotNull(converters.find { it.canConvert(input) }) {
+            "JsonConverter for '${input::class}' has not been registered"
+        }
+
+    fun <T: Any, JN : JsonNode> register(target: KClass<T>, converter: JsonConverter<T, JN>): KondorJson = apply {
```
daviddenton, line 48: In all of the other JSON modules, all of the converters are registered at construction time. Can we please make it match that pattern so we have consistency? 😄

daviddenton:
Looks mostly good - have left a couple of tweaks :)

http4k-format/kondor-json/src/main/kotlin/org/http4k/format/KondorJson.kt
FredNordin: No problem, will do that. Will probably mean I can hide some methods/classes that I has to expose publicly to make it work, which is a Good Thing :tm:

FredNordin: Done. I'm pretty happy with the outcome - much cleaner now and consistent with other format modules.

FredNordin:
Codecov isn't lying. There are some code paths that are not covered by the shared tests, but there are also a few methods in KondorJson for auto lenses that don't have test for all possible input params.

@daviddenton want me to have a look at adding some tests to up the coverage as part of this PR?

daviddenton:
That would be amazing - it’s pretty stable though so don’t go to too much effort :)

FredNordin:
I'm all done - ready for final review.

uberto (approved):

daviddenton:
I've added one small comment/question about the construction of the Kondor instance. If we sort that out then we can merge away 😄

uberto:
> I've added one small comment/question about the construction of the Kondor instance. If we sort that out then we can merge away smile

Sorry, it's Monday... but I cannot see the comment. Which file?

daviddenton:
KondorJson 🙃

uberto:
> I cannot see the comment either. GH having issues?

Yeah, there is no comment in KondorJson, as far as I can see

uberto:
> > > I cannot see the comment either. GH having issues?
> > 
> > 
> > Yeah, there is no comment in KondorJson, as far as I can see
> 
> https://github.com/http4k/http4k/pull/893/files#diff-01191e8c3685d3e35922d27c6de03ecb7974254284c28759bae11a7337c1d2a6 smile

I can see the file, but without any comment. On this PR there is only one conversation and it's marked as closed ;)

image

FredNordin:
> Then I'l post it here! smile
> 
> `Sorry for the further comment - can we make this match the other implementations which uses the asConfigurable() pattern? Want to be able to register particular converters for value types etc. if that makes sense in a Kondor world?`

@daviddenton So I did have that in a previous incarnation of the implementation, but I scrapped it because it didn't really makes sense in the KondorJson world. In KondorJson there is no concept of a "contextual" lookup, like there is in say Jackson or KotlinxSerialization where adapters for field types can be looked up at runtime, but rather converters for types are defined as singletons and referenced when creating other converters (see the test for good examples of this).
To "compensate" for this I have added methods to convert BiDiMapping classes to converters, which you can also see examples of in the test.

I can add some syntactic sugar to allow the user to start the initialisation process of KondorJson by using a "asConfigurable()" method, but I fear that might also confuse them, because it won't be a `AutoMappingConfiguration` type returned - there is no real value in being able to call `text()` or `int()` etc when the mappings done there are not able to be looked up.

The history of this PR contains that implementation if you really would like to see what it would look like, but neither @uberto nor me where very keen on it.

Not sure if any of this makes sense :thinking:

uberto:
The whole idea of Kondor is that once you have defined a Converter for type
ABC, you have it defined completely, so it knows how to convert ABC,
without resourcing to a registry for the "subtypes".
it may make sense to define a registry for the root types, so you can use
an auto lens (with inline refied sniffing). If I understand correctly this
is what Fred implemented.

daviddenton:
Ok. I'm in a bit of a dilemma now around this module. The latest version of Kondor requires Java 17, whereas we currently build on 11 and target 8.

We don't really want to break compatibility for J11 users by enforcing an upgrade to 17. I'm slightly worried that we will be stuck on Kondor 1.8.X until whenever that step is taken.

@uberto - is there a good reason that you're requiring 17 in Kondor?

uberto:
I created a branch of Kondor for this, that would stay on target J8, it can
stay there I suppose, I mean I don't plan any big improvements on
kondor-json any time soon. I'm already quite busy with the KondorMongo
thingy. And in case I can always backport them to kondor 1.8.x

Generally speaking I think It makes sense for a library to stick with the
latest LTS JDK, which is currently 17, but I'm not terribly opinionated
about it.

Uberto

daviddenton:
I know what you mean. We had the same thing, but decided that unless we had a good reason that we'd try and extend support for people as far back as possible so that Android and ah-I'm-trapped-in-a-crappy-environment teams could still use http4k. Currently the only thing we rely on from j9 onwards is the new version of the JavaHttpClient and we kept the other one around for others to use (Java8HttpClient). Also, having to maintain multiple branches and backport fixes seems like something we don't really want to be doing unless we have a reason to do it.

We're intending to upgrade the build JDK to 19/20 with v5, but keep the compatibility to 8.

daviddenton:
I've tried again with 1.9.2:

```

> Task :http4k-format-kondor-json:compileTestKotlin FAILED
e: file:///Users/david/dev/http4k/http4k/http4k-format/kondor-json/src/test/kotlin/org/http4k/format/KondorJsonAutoMarshallingJsonTest.kt:155:18 Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option
e: file:///Users/david/dev/http4k/http4k/http4k-format/kondor-json/src/test/kotlin/org/http4k/format/KondorJsonAutoMarshallingJsonTest.kt:156:20 Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option
e: file:///Users/david/dev/http4k/http4k/http4k-format/kondor-json/src/test/kotlin/org/http4k/format/KondorJsonAutoMarshallingJsonTest.kt:174:18 Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option
```
Not sure if we can work around this somehow with gradle magic? I'm happier for the test code to be built to a different target level - but my gradle-fu (ie. my patience)  isn't good enough to deal with this right now. If someone wants to take a stab at it then feel free 😄

uberto:
I can make another release for the 1.8 branch. I am not completely against
1.8 output but I suspect it would break projects which use later JVM,
considering the J9 changes...
I'll do a check.

Uberto

daviddenton:
What did you have in mind for those breakages? We use J11 to build currently and target 1.8 and never had problems in any project using 11-19.

FredNordin:
Hello :wave: So where do we stand now wrt this module? I can see that the module is still commented out in the `settings.gradle.kts` file, which I presume is because of the last few comments in this discussion.

uberto:
New version of Kondor (1.9.3) is jdk8 compatible, so it should be ok now.

daviddenton:
Great, we're scheduling it for the v5 release which is due to land this week.

Issue #912, opened 2023-05-25
Pass request to failResponseFn in ServerFilters.CatchLensFailure


mikaelstaldal:
It would be nice if `failResponseFn` in `ServerFilters.CatchLensFailure` get the request as an additional parameter, so you can return different type of error bodies depending on the `Accept` request header.

PR #913, opened 2023-05-25
Pass request to failResponseFn in ServerFilters.CatchLensFailure

mikaelstaldal:
Resolves https://github.com/http4k/http4k/issues/912

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 This list is not intended to be all-encompassing - it will document major and breaking API 
 changes with their rationale when appropriate:
 
+### v4.45.0.0 (uncut)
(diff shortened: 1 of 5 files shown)


daviddenton:
Thanks. A couple of bits:

1. I think we can maybe get away with the same name (and to just not have the default argument on the new function)
2. Can we cover the new function with a test?

😄

mikaelstaldal:
> 1. I think we can maybe get away with the same name (and to just not have the default argument on the new function)

It doesn't seem so, I get overload resolution ambiguity in `JsonRpcService`:

```CatchLensFailure { Response(OK).with(jsonLens of renderError(ParseError)) }```

mikaelstaldal:
> 2. Can we cover the new function with a test?

Yes, test added.

PR #914, opened 2023-05-26
adds the (proposed) 451 status code "Unavailable For Legal Reasons"

gypsydave5:
https://en.wikipedia.org/wiki/HTTP_451

I added it to the LiveTemplates as that seemed like the right thing to do, but not anywhere else


daviddenton:
Thanks! :)

PR #917, opened 2023-06-01
Truncate file resource timestamp to seconds. #916

### LINKED ISSUE #916: Experimental FileResource: File's last modified timestamp is not truncated to seconds when compared
https://github.com/http4k/http4k/blob/d56978c1d05ca9a28b90e617b9e5163a7f166284/http4k-incubator/src/main/kotlin/org/http4k/routing/experimental/FileResource.kt#L17

This line should truncate the file's last-modified timestamp to seconds (like `ResourceLoaders.ClassPath`), to avoid always returning 200s from the server.

What happens now is as follows:
- `file` is modified at 2022-06-01T18:00:00.388Z
- Client requests `file` and gets a response with `Last-Modified: 2022-06-01T18:00:00Z` (correctly)
- Client requests file again and sets `Last-Modified: 2022-06-01T18:00:00Z` (correctly)
- Server compares the request's `Last-Modified` with `2022-06-01T18:00:00.388Z`, and decides that the file was modified (incorrectly)

bjornbugge:
Closes #916

### DIFF
--- a/http4k-incubator/src/main/kotlin/org/http4k/routing/experimental/FileResource.kt
+++ b/http4k-incubator/src/main/kotlin/org/http4k/routing/experimental/FileResource.kt
@@ -4,6 +4,7 @@ import org.http4k.core.ContentType
 import java.io.File
 import java.io.InputStream
 import java.time.Instant
+import java.time.temporal.ChronoUnit.SECONDS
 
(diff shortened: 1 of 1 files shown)


daviddenton:
Thanks 🙃

PR #920, opened 2023-06-05
fix!: Status.from

jhult:
It should be inside the companion object


daviddenton:
I've fixed this in an unrelated commit, so will close it :)

PR #926, opened 2023-06-17
add option to mark contract routes as not described

### LINKED ISSUE #910: Allow RouteMetaDsl to be marked as hidden
The `RouteMetaDsl.markAsDeprecated` method is useful, but sometimes I'd like to go the extra mile and actually exclude it from the OpenApi spec.  This can be done by converting the `ContractRoute` to a regular non-contract `HttpHandler`, but it would be nice to be able to perform a deprecation with minimal effort, and still take advantage of the path routing.

I propose adding a `RouteMetaDsl.markAsHidden` method that will prevent it from being included in the OpenApi spec.  Thoughts?

oharaandrew314:
Closes #910


daviddenton:
👍

PR #927, opened 2023-06-17
Gracefully handle JavaWebSocket events when http4k adapter not yet registered

oharaandrew314:
If any WebSocket events are sent (such as a close) during the `WsHandler` invocation, it would result in an NPE and potentially kill the server.  This fix attempts to tackle the problem from two directions:

1. Register the `PushPullAdaptingWebSocket` before processing the upgrade request
2. If the `PushPullAdaptingWebSocket` hasn't been registered, gracefully avoid an NPE


daviddenton:
👍

PR #934, opened 2023-06-23
Remove note regarding using the http4k implementation of By as it was…

chighfield-sn:
Remove note regarding importing the http4k implementation of `By` (`org.http4k.webdriver.By`) as it was deleted in the following commit: [https://github.com/http4k/http4k/commit/974a582f79f066bde6ef93945eb20c66915a81a0](https://github.com/http4k/http4k/commit/974a582f79f066bde6ef93945eb20c66915a81a0)


s4nchez:
Thank you for spotting it, @chighfield-sn !

PR #935, opened 2023-06-23
Add custom json marshaller for AccessTokenRequest

### LINKED ISSUE #929: http4k-security-oauth requires reflection
http4k-security-oauth relies on kotlin-reflect to be present on the classpath to (de)serialize token requests and responses.  This isn't ideal on serverless environments, where kotlin-reflect is avoided.

I propose we eliminate `kotlin-reflect` from `http4k-security-oauth`, and replace it with hand-crafted Moshi serializers like in `http4k-serverless-lambda:AwsLambdaMoshi`, or with kotshi serializers as is done in http4k-connect.

oharaandrew314:
This could have been done with minimal churn, since the only thing we were missing was for the `AccessTokenRequest`.

But since the original Moshi instance was called `OAuthServerMoshi`, yet it applies to the client too, I've renamed and repackaged it, and have modernized the existing custom marshallers.

Closes #929


oharaandrew314:
Interestingly, the oauth module already excludes kotlin-reflect.  I believe it was simply an oversight to introduce `TokenRequest` with a dependency on it.

daviddenton:
thanks! I'm going to move the Moshis back below the security package, but other than that 👍

PR #938, opened 2023-06-29
Request with UriTemplate parameters

potfur:
There's an easy way of instantiate `RoutedRequest` with `UriTemplate`, but not with applied properties.

This MR adds this as another `invoke` to `Request`

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/http.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/http.kt
@@ -259,6 +259,13 @@ interface Request : HttpMessage {
 
         operator fun invoke(method: Method, template: UriTemplate, version: String = HTTP_1_1): Request =
             RoutedRequest(Request(method, template.toString(), version), template)
+
+        operator fun invoke(
+            method: Method,
+            template: UriTemplate,
+            parameters: Map<String, String>,
+            version: String = HTTP_1_1
+        ): Request = RoutedRequest(Request(method, template.generate(parameters), version), template)
     }
 }
 


s4nchez:
@potfur can you please include a test so we can demonstrate the behaviour and ensure things keep working in the future?

s4nchez:
Also, we'd be interested to understand the use case (i.e. when do you need to create a specific RoutedRequest directly with the parameters applied?)

daviddenton:
You can already do this type of thing in a more typesafe way with. TBH I'm a bit adverse to encouraging people to just passing maps of strings around :) 

```kotlin
    val message = Request(GET, UriTemplate.from("/{string}/{int}"))
            .with(Path.of("string") of "hello", Path.int().of("int") of 1)
    println(message)
```

potfur:
The use case is - create a `Request` instance that has URI with values and the template used for generating that URI.
With this it is possible to write a filter or a handler that will report issue on endpoint `/foo/{id}` instead of `/foo/123` - better error grouping.

But I see that the way shown by @daviddenton would solve this problem without introducing _kind of weak_ factory function - so, PR is being closed. 😄

daviddenton:
Yep. It's definitely a worthy case - so we can report the template in metrics etc. There is a bit of a user-tax for sure as you need to declare the template and then fill it in, but I do think the consistency is worth it for the lens type-safety. 

Maybe we're missing docs on this... 🤔

potfur:
> Maybe we're missing docs on this... 🤔

Yup, that may be the thing that's missing.
Finding `with(Path ... ` when you know how this works is easy, but just following `with(...)` didn't give me any helpful results.

PR #939, opened 2023-06-30
Override all `Request` and `Response` mutators for routed messages to…

kwydler:
… ensure proper return type.

Looks like several mutator methods in `RoutedRequest` and `RoutedResponse` are not overridden so they are returning the delegate message type.

Example:
```kotlin
import org.http4k.core.Method
import org.http4k.core.Request
import org.http4k.core.UriTemplate
import org.http4k.core.with
import org.http4k.lens.Path
import org.http4k.lens.Query
import org.http4k.lens.int
import org.http4k.lens.string

fun main() {
  val name = Path.string().of("name")
  val size = Query.int().optional("size")

  val template = UriTemplate.from("/foo/{name}")

  val request = Request(Method.GET, template) // RoutedRequest
    .with(name of "bar") // RoutedRequest
    .with(size of 1) // MemoryRequest

  print(request)
}
```

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/routing/messages.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/routing/messages.kt
@@ -4,7 +4,9 @@ import org.http4k.core.Body
 import org.http4k.core.Headers
 import org.http4k.core.Method
 import org.http4k.core.Request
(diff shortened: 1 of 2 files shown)


daviddenton:
Good catch! Thanks :)

PR #944, opened 2023-07-07
Reword module description

Alfhir:
In the original version, the enumerated list wasn't rendered correctly (see discussion on [slack](https://kotlinlang.slack.com/archives/C5AL3AKUY/p1688642812413349)).
I propose to name the features this module provides after stating that there are two of them for better readability.


s4nchez:
Thank you for the contribution, @Alfhir !

PR #946, opened 2023-07-17
Scan all constructors for parameter descriptions

### LINKED ISSUE #750: JacksonFieldMetadataRetrievalStrategy is incompatible with kotlinx.serialization @Serializable classes
Introduced in: #467 #687 

The implementation uses the first constructor found:
https://github.com/http4k/http4k/blob/115d4cb0ff3643bd1e2fec41ecac3806431b0d80/http4k-contract/src/main/kotlin/org/http4k/contract/openapi/v3/jacksonExt.kt#L82

However, a class annotated with `@Serializable` gets an extra constructor, and the code picks this one first.

```kotlin
@kotlinx.serialization.Serializable
data class ExampleDto(
    // This annotation DOES NOT WORK if you use kotlinx.serialization @Serializable on this class.
    @JsonPropertyDescription("The unique id for this entity")
    val id: String
)
```

REPL output:
```kotlin
 val size = ExampleDto::class.constructors.size
 val metadata = JacksonFieldMetadataRetrievalStrategy.invoke(ExampleDto("1"), "id")
 println("There are $size constructors and the metadata is ${metadata.extra}")
```

`There are 2 constructors and the metadata is {}`

If I remove only the `@Serializable` annotation and run the same code, I get this output:
`There are 1 constructors and the metadata is {description=The unique id for this entity}`

krissrex:
Fixes #750

## Background

When using `kotlinx.serialization` `@Serializable` for data classes in conjuction with the Jackson OpenApi generator, http4k will fail to pick up annotations on data classes.

Here is an example data class with `@Serializable`:
image

It will generate/compile code like this:
image
image

That `Deprecated` constructor is a hidden and generated constructor.
When http4k tries to find annotations, this one is being selected instead of the original constructor.

## Fix

Scan all constructors instead of just the first.
I used sequence, so we don't crawl all constructors needlessly, it should only go one-by-one until a description is found.

A test case has been added.
I use a class like this:
image

which tries to emulate the `@Serializable` class.
The compiled code looks like 
image
.

## Lint

I didn't find any resources or commands for linting this project.
Thus, code may be formatted incorrectly.

Just edit any code or remove comments if you dislike them.

### DIFF
--- a/http4k-contract/jsonschema/src/main/kotlin/org/http4k/contract/jsonschema/v3/jacksonExt.kt
+++ b/http4k-contract/jsonschema/src/main/kotlin/org/http4k/contract/jsonschema/v3/jacksonExt.kt
@@ -55,12 +55,19 @@ object JacksonFieldMetadataRetrievalStrategy : FieldMetadataRetrievalStrategy {
         FieldMetadata(target.javaClass.findPropertyDescription(fieldName)?.let { mapOf("description" to it) }
             ?: emptyMap())
 
+    /**
+     * Scan all constructors until one contains the property named [name] and has an [JsonPropertyDescription]
+     * annotation with non-null value.
+     *
+     * By scanning multiple constructors, this also works in cases with generated constructors and no-arg constructors.
+     */
     private fun Class<Any>.findPropertyDescription(name: String): String? =
-        kotlin.constructors.firstOrNull()?.let {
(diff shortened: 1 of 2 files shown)


daviddenton:
Thanks for the contrib. Will get this merged and released ASAP :)

PR #948, opened 2023-07-17
Fix bug where spaces are not deduplicated. Merge duplicated headers

### LINKED ISSUE #947: AWS signature converts headers incorrectly to canonical headers
Related to #656 .

Two bugs: spaces are not deduplicated, and duplicate header values are not joined.

I have a fix for both coming up, for headers. 
Not a fix for the object key thing, in #656.

The spaces are not deduplicated because the `"\\s+"` regex was passed as a literal string pattern.

krissrex:
Fixes #947

https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html#create-canonical-request

The duplicate header issue is solved by grouping header values before sorting.
I turn it to a `Map<Header key, List<Header value>>` before sorting headers, then join the list of values with `,`.


s4nchez:
Thank you for spotting + fixing this 😄

Issue #958, opened 2023-08-07
RouterBasedHttpHandler is internal


geertmulders:
Normally I add a list of routers by using the `routes(vararg list: RoutingHttpHandler)` function. Now I wanted a custom error handler is the route could not be found; luckily behind `routes` there is the `RouterBasedHttpHandler` class, that can do just that!

Less luckily this class is internal, so I cannot use it.

Adding the following would really help:

```
    private fun routes(
        vararg list: RoutingHttpHandler,
        notFoundHandler: HttpHandler = routeNotFoundHandler,
        methodNotAllowedHandler: HttpHandler = routeMethodNotAllowedHandler,
    ): RoutingHttpHandler =
        routes(list.toList(), notFoundHandler, methodNotAllowedHandler)

    private fun routes(
        list: List<RoutingHttpHandler>,
        notFoundHandler: HttpHandler = routeNotFoundHandler,
        methodNotAllowedHandler: HttpHandler = routeMethodNotAllowedHandler,
    ): RoutingHttpHandler =
        RouterBasedHttpHandler(OrRouter.from(list), notFoundHandler, methodNotAllowedHandler)
```

I can make a PR if you like.

daviddenton:
You can actually already do this without needing to expose any more complexity into the routing: 

```kotlin
val myApp = routes(
    "/bob" bind GET to { _: Request -> Response(OK) },
    Router.orElse bind { _: Request -> Response(NOT_FOUND) },
)
```

daviddenton:
Closing this off now as don't think it's required 😄

PR #959, opened 2023-08-07
Routing with error handlers #958

geertmulders:
This resolves issue #958


jhult:
@geertmulders, it [sounds like](https://github.com/http4k/http4k/issues/958#issuecomment-1667540730) this isn't necessary. Should this be closed?

daviddenton:
@jhult Agreed - I'll close this off for now. We can reopen if there is reason to. Thanks everyone for the input! 😄

PR #972, opened 2023-09-01
Support supplying a custom worker group to Netty server

mikaelstaldal:


daviddenton:
Sorry - we won't merge this. We made a decision a long time ago to not have custom options be passed to the ServerConfig constructors. If you need something like this then the official guidance is to create your own version of the ServerConfig - and we've made things as modular as possible to enable this.

🙃

PR #981, opened 2023-10-05
Update documentation to reflect Pebble model binding

cosmin-marginean:
This PR addresses the documentation changes to reflect the fact that Pebble templates will require a `model.` prefix to access the underlying model for the view.


daviddenton:
Thanks for the investigation! 🙃

PR #982, opened 2023-10-08
Duplicated the classes from the underscore-java library

javadev:


javadev:
🌞

daviddenton:
I'm a bit lost - what was the intended purpose of this change? 

It seems like we've just replaced a dependency with copying in some classes from that lib which we didn't write and would therefore need to maintain.

javadev:
The classes are currently quite stable. The Kotlin classes seem to be more suitable for the library. It's worth noting that the library now has a dependency on Java 11.

javadev:
➕

s4nchez:
@javadev we're inclined to not include this change. 

As much as we want to minimize our dependencies, this module is meant to be used for testing purposes and it's hard to justify adding thousands of lines of code to our module, even though they may be stable.

s4nchez:
We'll close this for now but can continue the discussion if needed.

PR #998, opened 2023-11-06
Allow the originalUri method of the OAuthRedirectionFilter to be conf…

mbcltd:
…igured when constructing an OAuthProvider.

Right now you can't configure the originalUri which prevents capturing OAuth URLs from domains other than the domain on which the OAuth controller is hosted.

### DIFF
--- a/http4k-security/oauth/src/main/kotlin/org/http4k/security/OAuthProvider.kt
+++ b/http4k-security/oauth/src/main/kotlin/org/http4k/security/OAuthProvider.kt
@@ -1,6 +1,7 @@
 package org.http4k.security
 
 import org.http4k.core.HttpHandler
+import org.http4k.core.Request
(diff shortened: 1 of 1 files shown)


mbcltd:
I don't think that my original description was very clear, this is needed in order for a process to manage authentication across multiple domains, e.g.:

- If the auth callback endpoint is on https://auth.foo.com/callback
- But the resource protected by the authfilter is on https://content.foo.com/something
- Currently the OAuthProvider will only call the OAuthPersistence with the path `/something` and not the full URI
- The OAuthPersistence, however, needs the full URI in order to redirect back to return the full correct redirect Uri when `retrieveOriginalUri(...)` is called

This change allows the user of OAuthProvider to optionally pass in their own `originalUri` function which can extract the full Uri including domain from the request.

PR #999, opened 2023-11-08
Make EnvironmentKey.secret a BiDiLens

oharaandrew314:
Given

```kotlin
val secretKey = EnvironmentKey.secret().required("PASS")
val fooKey = EnvironmentKey.required("FOO")
```

When building a test environment, I want to be able to use

```kotlin
val env = Environment.defaults(
    secretKey of Secret("hunter2"),
    fooKey of "bar"
)
```

but the best I can do is

```kotlin
val env = Enviromment.from(
    secretKey.meta.name of "hunter2"
) overrides Environment.defaults(
    fooKey of "bar"
)
```

So I've updated `EnvironmentKey.secret` to return a `BiDiLens`, which gives the functionality I'm looking for.  However, I wonder if I'm approaching this from the wrong direction.  If `EnvironmentKey.secret` returns a `Lens`, then I don't see how my intended usage doesn't work out of the box.  Perhaps I lack sufficient understanding of the internals of the lens system.  Is what I've done the right approach?

### DIFF
--- a/http4k-cloudnative/src/main/kotlin/org/http4k/cloudnative/env/Authority.kt
+++ b/http4k-cloudnative/src/main/kotlin/org/http4k/cloudnative/env/Authority.kt
@@ -1,5 +1,7 @@
 package org.http4k.cloudnative.env
 
+import org.http4k.core.Uri
+
 data class Authority(val host: Host, val port: Port? = null) {
     override fun toString() = host.value + (port?.let { ":${it.value}" } ?: "")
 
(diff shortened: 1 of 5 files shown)


http4k-cloudnative/src/main/kotlin/org/http4k/cloudnative/env/Authority.kt
```diff
@@ -13,3 +15,6 @@ data class Authority(val host: Host, val port: Port? = null) {
         }
     }
 }
+
+fun Authority.asHttpsUri() = Uri(scheme = "https", userInfo = "", port = port?.value, host = host.value, path = "", query = "", fragment = "")
```
oharaandrew314, line 19: This is unrelated, but I think it can be useful.

PR #1000, opened 2023-11-09
Support redirects in Http4kWebDriver's emulation of a client's host header

mbcltd:
Currently redirect chains to a different host will all have the host header of the domain that triggered the first redirect, not the true host header that would be seen from a browser following the redirects.


daviddenton:
have you checked out the changes to followredirects - is this related to that?

mbcltd:
No, this would only aver set the host header on Navigate.to and would have ignored follow redirects

mbcltd:
The issue was down to the order of the request filters on Http4kWebdriver, we were adding the host header to the request **after** following redirects, not before.

Issue #984, opened 2023-10-09
Can't upgrade Jetty version


sofiaspringer:
Hello, the project I am working on depends both on http4k and jetty (which as you know is also a dependency of http4k). However, we are blocked from upgrading Jetty version to 12 because http4k most recent version still uses Jetty 11. Is there a new version of http4k using Jetty 12 being launched soon? 

Many Thanks

daviddenton:
Unfortunately, Jetty 12 is almost a total rewrite of the Jetty 11 APIs and thus will require significant rework of the http4k code as well. Also, the migration docs aren't very informative (happy to be wrong on this if people have pointers!).

We do know about it but would appreciate any help to get the APIs migrated - it's definitely not a priority for us (especially because we personally tend to use Undertow as our server of choice :) )

daviddenton:
it's what we did for apache 4-5, so that seems ok - however for that version we actually made the OLD one the versioned one and left new one to be without the number. I'd propose the same thing here, so we'd have:

http4k-server-jetty-11 (v11)
http4k-server-jetty (v12)

FredNordin:
What happens when Jetty 13 is released? :grin:

FredNordin:
Seriously though, the naming of the new modules is less of an issue and I would always favour consistency.

daviddenton:
As long as the APIs are compatible, we've generally upgraded major versions silently (with a Semver bump of course) as it doesn't really affect the vast majority of users. eg. we went from jetty 8 -> 11 with no issues :)

FredNordin:
I'm tempted to have a go... (famous last words)

FredNordin:
So the module name of `org.eclipse.jetty:jetty-server` is still the same in version 12, which mean we won't be able to (AFAIK) have two versions defined in the `versions.properties` file for `refreshVersions` plugin. That means, unless we do something hacky, that having two modules targeting different versions of Jetty would be difficult.
Maybe just have to bite the bullet and upgrade Jetty to 12...

daviddenton:
We might be able to work around that, so having a decent idea about the work involved in an upgrade would be helpful. J21 is out now, so (as soon as Helidon upgrades to it - the preview features flags from v20 are incompatible with v21! 🙄 ) we can actually bump the major version number, which would be a good time to upgrade to Jetty12 as well :)

FredNordin:
[...]

PR #1003, opened 2023-11-14
Upgrade to Jetty 12

FredNordin:
This PR addresses the issue raised in #984 

The changes made are definitely breaking backwards compatibility and thus requires a major version upgrade of http4k.

A side-effect of upgrading to v12 is that the Jetty server code is no longer dependent on Servlet libs. I have left the support for converting http4k Request/Response objects to Servlet equivalent objects in the core, as the decision to leave it in or take it out is not within the scope of this PR.

I have tried to keep the existing code as much as possible, but because of the major change to the Jetty API there are quite a few differences in the internal implementation classes. I have followed the programming guide instructions as much as possible though for the new implementation bits.

Jetty server now also supports StopMode.Immediate, which it didn't previously.

I've also changed all the test classes that start/stop a Jetty server instance to use StopMode.Immediate instead of StopMode.Graceful. The reason is that Jetty, for some strange reason, take about 2 seconds to gracefully shut down. That happens for each test method, which quickly adds up in terms of total build time. For example, I shaved 20s off the total time on my laptop when switching to Immediate.
This is not a high priority change though, so if you'd rather keep it the way it was and are happy to the build to take longer to complete, I'm happy to revert those particular changes.


daviddenton:
Thanks v much @FredNordin for all the work!  We'll take this away and decide how to get it in soon. You're right - it will involve a major version upgrade. The good news is that people can still use the old version (maybe we'll make the old one http4k-server-jetty11 and have this supercede it. Am also very keen to see how it performs in the benchmarks - although we have just missed round 22 of TechEmpower :( 

🙃

PR #1004, opened 2023-11-15
Add an extension on Response to mark it as Cache-Control: immutable

moddular:
A small extension to mark a response as Cache-Control: immutable, useful for revved assets with long cache times.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control


daviddenton:
ta!

PR #1008, opened 2023-12-12
Allow changing the host option for Undertow

raphiz:
The current implementation of the Undertow Http4kServer always binds itself to `0.0.0.0`. This binds it to _all_ IPv4-enabled network interfaces. 

Binding to _all_ IPv4-enabled network interfaces is often not a good idea from a security point of view. For example, binding to local interfaces only (`127.0.0.1`) is usually sufficient for local development and does not expose the application on public network interfaces.

It's also not sufficient when IPv6 is desired.

This change simply adds the option to override the binding host as desired  - for undertow only. I have kept the change as minimal as possible to avoid introducing a breaking change. One could also think about whether the default should be adjusted.

### DIFF
--- a/http4k-server/undertow/src/main/kotlin/org/http4k/server/Undertow.kt
+++ b/http4k-server/undertow/src/main/kotlin/org/http4k/server/Undertow.kt
@@ -17,7 +17,8 @@ import java.net.InetSocketAddress
 class Undertow(
     val port: Int = 8000,
     val enableHttp2: Boolean,
-    override val stopMode: StopMode = StopMode.Immediate
+    override val stopMode: StopMode = StopMode.Immediate,
+    val host: String = "0.0.0.0",
 ) : PolyServerConfig {
     constructor(port: Int = 8000) : this(port, false)
     constructor(port: Int = 8000, enableHttp2: Boolean) : this(port, enableHttp2, StopMode.Immediate)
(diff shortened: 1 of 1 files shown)


daviddenton:
Thanks for taking the effort to report this. We've had a number of suggestions like this and always consider them, but mostly to avoid an explosion of configuration options, we fallback to our decision we took a while ago.. From the [site](https://www.http4k.org/guide/reference/servers/#customisation): 

```
Each of the server backends implement an interface ServerConfig, which is written with sensible defaults for the server in questions, but is also designed to be used as a starting point for tweaking to API user needs. To customize, simply use the relevant ServerConfig class as a starting point and reimplement as required. See the how-to guides for an example of this in use.
```

😄

raphiz:
Sure, that makes sense!

I'm not in favour of many configuration options at all, especially since this interface is so simple and easy to adapt.

However, I do think that just providing a configuration for the port and not the host is somewhat incomplete. The tuple of address and port is the lowest common denominator of all webserver configurations.

In any case, I trust your judgment as a maintainer of this great project - it is important to be able to say no to certain requests. It's fine to close this issue if you disagree.

As a side note: Would it make sense to consolidate all `Server` implementations to be consistent regarding this issue? For example the `ApacheServer` supports setting an address and `UndertowServer` has an additional `enableHttp2` option.

daviddenton:
You're completely right - we have talked about the removal of these other options that have crept in over time to make this inconsistent, as making things simple is the most important thing for our users. We'll probably look at doing it for the next major release (we need to merge in the breaking upgrade to V12 of jetty anyway so this would be a good opportunity!)

Thanks v much for the kind words. 🙃

daviddenton:
I've raised this to track the issue. Thanks again

https://github.com/http4k/http4k/issues/1010

PR #1014, opened 2023-12-20
Aws presigner

oharaandrew314:


http4k-aws/src/main/kotlin/org/http4k/aws/AwsPreSignRequests.kt
```diff
@@ -0,0 +1,39 @@
...
+    private val scope: AwsCredentialScope,
+    private val credentials: AwsCredentials,
+    private val clock: Clock = Clock.systemDefaultZone()
+) {
+    operator fun invoke(request: Request, expires: Duration): AwsPreSignedRequest {
+        val time = clock.instant()
+        val awsDate = AwsRequestDate.of(time)
+
+        val headers = request.headers + ("host" to request.uri.host)
+        val signedHeaders = headers.map { it.first.lowercase(Locale.getDefault()) }.toSet().sorted().joinToString(";")
```
oharaandrew314, line 19: I could get this from the `AwsCanonicalRequest`, but I need these to make the `AwsCanonicalRequest`.  🤷 

Worth extracting the common code?

daviddenton: Yes - let's try to isolate it, even if it's an internal scope :)

daviddenton:
Is it worth making this entire thing into a Filter

http4k-aws/src/main/kotlin/org/http4k/aws/AwsPreSignRequests.kt
oharaandrew314: Done

oharaandrew314:
>Is it worth making this entire thing into a Filter?
Don't think so.  The whole idea is to share the pre-signed URI with a client, so making this into a `ClientFilter`--while possible--is unnecessary.  Or is there another use-case I'm not considering?

Ready for review again.

PR #1015, opened 2023-12-20
aws presigner to accept credentials provider

oharaandrew314:
I forgot to support a credentials provider.  Here it is.

I think I know how @daviddenton would have architected this. Rather than have a class with multiple constructors, there would have been an `object AwsPresignRequests` with two `operator fun invoke` functions (or just two top-level `AwsPresignedRequest` functions) that return a `(Request, Duration) -> AwsPresignedRequest`.

It makes no real difference to the user, but if you would prefer I stick to the established http4k style, I don't mind changing it.

### DIFF
--- a/http4k-aws/src/main/kotlin/org/http4k/aws/AwsPreSignRequests.kt
+++ b/http4k-aws/src/main/kotlin/org/http4k/aws/AwsPreSignRequests.kt
@@ -5,42 +5,44 @@ import org.http4k.filter.Payload
 import java.time.Clock
 import java.time.Duration
 
-class AwsPreSignRequests(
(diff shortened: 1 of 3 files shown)


daviddenton:
@oharaandrew314 close, but I'm glad I not yet that predictable - my job is safe for now - I went with the one below! Thanks for the contribution again. 😄 

```kotlin
fun interface AwsPreRequestSigner : (Request, Duration) -> AwsPreSignedRequest

fun AwsPreRequestSigner(
    scope: AwsCredentialScope,
    credentialsProvider: () -> AwsCredentials,
    clock: Clock = Clock.systemUTC()
): AwsPreRequestSigner
```

PR #1021, opened 2024-01-02
Include method in AwsPresignedRequest

oharaandrew314:


daviddenton:
is this preventing the signing from working? (ie. is it a bug?)

oharaandrew314:
No, including the signed method in the result is for convenience only.

The only reason the signature of one of the tests changed was because I changed the method used in the test.

Issue #538, opened 2020-12-04
JWKs support


ikovalyov:
Hi.
We are using the jwks in our application to authenticate customers. Unfortunately I can see only the OAuth security module. Do you have any plans on adding the JWKs support?
Thanks.

daviddenton:
TBH we have no plans to add a module like this at the moment. PRs gratefully accepted though ;)

What would be interesting is to find out exactly what type of functionality our users would like to see from such a module, so please feel free to enumerate stuff in this thread. 😄

razvn:
It's quite easy to use JWT with current implementation of BearAuth then it's up to you to handle the JWT. Plus it gives you the liberty to use the JWT lib of your choice.
Maybe the only thing missing is a recipe in the Cookbook which could show an example of how to use it. I can't see what a higher abstraction of a JWT, http4k could add.

ikovalyov:
@razvn I'd like to have a module which has something like this:

- We can configure the jwks endpoint (file which lists the issues keys)
- We can configure the structure and types of the issues keys (algo, encryption and other stuff)
- The system adds a firewall which will decode the jwt, verify it over well-known keys and check if user has a permission to access the endpoint
- System uses cache to store list of the well known keys for future use.

razvn:
I agree it would be nice having that included in the library, instead of having to do it manually as I had to do a few weeks ago. 
But I think is quite specific usage, as I guess most JWT does not use JWKS severs. 
Beside it requires some asynchronous tasks (refreshing the JWKS in the background) which the library does not handle yet (I know it can be done with ComputableFutures (I went with coroutines as I like them better)) which adds complexity and difficulty to make it configurable to fit each one requirement.

oharaandrew314:
Sorry to necro this, but a use-case for integrating with http4k did come to mind recently.  I was originally using [auth0 java-jwt](https://github.com/auth0/java-jwt), but then for jar size limitations, I wanted to eliminate guava and jackson.  I ended up using [Nimbus Jwt](https://connect2id.com/products/nimbus-jose-jwt), because it uses a low-level json library; but I did have to give up some of the goodies from auth0's implementation.

I still don't know how we could do a better job than auth0, but it would be nice if we could bring our own http and marshalling backends.  Maybe if we include auth0 as an implementation dependency, exclude all its dependencies, and then provide a http4k-esque wrapper on top?

oharaandrew314:
[...]

PR #1024, opened 2024-01-08
New Module: http4k-security-jwt

oharaandrew314:
Add a new JWT security module based on `nimbus-jose-jwt`, including:

- Server filters
- Contract Security
- JWK rendering
- Testability tools

Relates to #538


oharaandrew314:
The docs module can't properly import classes from the new module, but I'm not sure what I'm missing to make it available.  I thought the top-level build.gradle was supposed to handle it automatically.

daviddenton:
I'm going to merge this into secondary branch. We already had some thoughts about this type of thing (and have implemented similar things) so it would be worth aligning on those before it hitting master. Hope that's ok 🙃

oharaandrew314:
Ok sure.  Keep me in the loop.

Issue #1023, opened 2024-01-05
Header values in quotes lose their quotes


efasel:
[ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and [If-None-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) are headers where the values are quoted:
> It is a string of ASCII characters placed between double quotes, like "675af34563dc-tr34"

These quotes are lost here as the `headers.getCSV` method is called with `keepQuotes = false`: 
https://github.com/http4k/http4k/blob/457b8d70053454d39b5d9146c1a2f859820a1b0e/http4k-server/jetty/src/main/kotlin/org/http4k/server/Http4kJettyHttpHandler.kt#L35-L38

This later leads to a mismatch when comparing the if-none-match header value with the generated etag value.

This prevents us from upgrading to Jetty 12.

Co-Found-By @dhs3000

efasel:
```kotlin
Subject: [PATCH] Add test for header values with quotes
---
Index: http4k-server/jetty/src/test/kotlin/org/http4k/server/Http4kJettyHttpHandlerTest.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/http4k-server/jetty/src/test/kotlin/org/http4k/server/Http4kJettyHttpHandlerTest.kt b/http4k-server/jetty/src/test/kotlin/org/http4k/server/Http4kJettyHttpHandlerTest.kt
new file mode 100644
--- /dev/null	(date 1704475064584)
+++ b/http4k-server/jetty/src/test/kotlin/org/http4k/server/Http4kJettyHttpHandlerTest.kt	(date 1704475064584)
@@ -0,0 +1,322 @@
+package org.http4k.server
+
+import com.natpryce.hamkrest.assertion.assertThat
+import com.natpryce.hamkrest.equalTo
+import org.eclipse.jetty.http.HttpField
+import org.eclipse.jetty.http.HttpFields
+import org.eclipse.jetty.http.HttpURI
+import org.eclipse.jetty.http.HttpVersion
+import org.eclipse.jetty.io.Connection
+import org.eclipse.jetty.io.Content
+import org.eclipse.jetty.server.Components
+import org.eclipse.jetty.server.ConnectionMetaData
+import org.eclipse.jetty.server.Connector
+import org.eclipse.jetty.server.Context
[...]

PR #1027, opened 2024-01-08
Keep quoted headers quoted

efasel:
Fixes http4k/http4k#1023

[There is a test for this issue in the corresponding issue](https://github.com/http4k/http4k/issues/1023#issuecomment-1879101953), but I guess this can be done much simpler for someone who is much more familiar with http4k and Jetty.

```
BUILD SUCCESSFUL in 4m 14s
622 actionable tasks: 622 executed
```


efasel:
> Do you plan to merge the test into this branch?

Should I? I was hoping that the test could be done much simpler for someone who is much more familiar with http4k and Jetty.

dhs3000:
I've added a test based on your (@efasel) patch from the issue, using fakes as I see it elsewhere in these tests

dhs3000:
I moved the test to the ServerContract so that it checks/ensures this behaviour for all servers

PR #1033, opened 2024-01-13
Fix handlebars import in htmx example

Smponi:
The current import of the handlebars template engine in the htmx example is not working. (Probably outdated?)
This fixes that tiny bug.


daviddenton:
👍🙃

PR #1051, opened 2024-02-03
Update Swagger UI Config

oharaandrew314:
- Allow additional unstructured properties to be set
- Don't override UI defaults to `null`


http4k-contract/openapi/src/main/kotlin/org/http4k/contract/ui/MapDelegate.kt
```diff
@@ -0,0 +1,25 @@
+package org.http4k.contract.ui
+
+import kotlin.reflect.KProperty
+
+internal class MapDelegate<T>(
+    private val properties: MutableMap<String, UiProperty<out Any>>,
+    private val name: String,
```
daviddenton, line 7: do we need this value? the name does come in through the get/set value methods (as a part of the property) so we can probably get away without it?

oharaandrew314: THe only place we need it is for `dom_id`, unless we're ok breaking backwards compatibility by changing the property name to match.

oharaandrew314:
Tests are finally passing 😅

PR #1055, opened 2024-02-07
Host header should contain host with port

obecker:
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

Required for https://github.com/http4k/http4k-connect/issues/361


http4k-core/src/main/kotlin/org/http4k/core/Uri.kt
```diff
@@ -30,12 +30,16 @@ data class Uri(val scheme: String, val userInfo: String, val host: String, val p
...
-        .appendIfNotBlank(userInfo, userInfo, "@")
+    val hostAndPort = StringBuilder()
         .appendIfNotBlank(host, host)
         .appendIfPresent(port, ":", port.toString())
         .toString()
 
+    val authority = StringBuilder()
+        .appendIfNotBlank(userInfo, userInfo, "@")
+        .append(hostAndPort)
+        .toString()
```
obecker, line 41: I considered making them `lazy`, but I suppose it's not a big difference performance-wise.

daviddenton:
We've had a chat and decided to keep the hostAndPort property out of the Uri class. This does involve recreating duplication, but it's targeted duplication so we're ok with it for now and we now feel happier to merge it in to unblock the problem in connect. Thanks for the contrib again! We'll get both out ASAP 😄

http4k-core/src/main/kotlin/org/http4k/filter/ClientFilters.kt
```diff
@@ -182,7 +182,7 @@ object ClientFilters {
             val redirect = ensureValidMethodForRedirect().uri(newUri)
 
             return when {
-                header("host") != null && newUri.host.isNotEmpty() -> redirect.replaceHeader("host", newUri.hostAndPort)
+                header("host") != null && newUri.host.isNotEmpty() -> redirect.replaceHeader("host", newUri.host)
```
obecker, line 185: What about this one? Here a `host` header is set, so shouldn't it include the `port` as well?

obecker:
You could create an extension function for `hostAndPort` to keep it out of the `Uri` class ...

http4k-core/src/main/kotlin/org/http4k/filter/ClientFilters.kt
daviddenton: Missed that one. Will clean it up for the next release 🙃

PR #1058, opened 2024-02-12
Allow user to provide schema creation implementation

tamj0rd2:
I'm using a library called Kondor which provides a way to generate json schemas without the use of reflection etc. In the http4k OpenApi3 implementation, it's currently not possible to provide your own JsonToJsonSchema implementation, however it is in the OpenApi2 implementation.


daviddenton:
Interesting - is that schema generator open source by any chance? (or is it even a part of Kondor?)

daviddenton:
Thanks for the PR. It would be amazing if the `JsonSchemaCreator` implementation could be open sourced (either as a part of Kondor (ask @uberto  :)) or as a part of this module. There is a test with a bunch of test cases which could be useful in `AutoJsonToJsonSchemaTest`.

PR #1061, opened 2024-02-19
Symmetric ws handler

### LINKED ISSUE #1049: Support WsHandler symmetry
The `HttpHandler` supports perfect symmetry, where it can be used as the interface for a server and client.

Currently, the `WsHandler` is only supported on the server-side, making it more of a chore to fake websocket connections.

**Workaround**
```kotlin
class MyService(
    private val openWs: (Uri) -> Websocket 
) {
  fun open(): Websocket {
      return openWs(Uri.of("ws://foo"))
  }
}

val realService = MyService { uri ->
    WebsocketClient.nonBlocking(uri)
}

val testWs = websockets()
val testService = MyService { uri ->
    val request = org.http4k.core.Request(Method.GET, uri)
    testWs.testWebsocket(request)
}
```

**Desired**
```kotlin
class MyService2(
    private val wsHandler: WsHandler
) {
    fun open(): Websocket {
        val request = Request(Method.GET, "ws://foo")
        return wsHandler(request).nonBlocking()
    }
}

val realService2 = MyService2(JavaWebSocketClient())

val testWs2 = websockets()
val testService2 = MyService2(testWs2) 
```

Does this seem reasonable and doable?

oharaandrew314:
Closes #1049

You can finally inject a `SymmetricWsHandler` into your application and have it be implemented by a server or a client.  This will make testing websocket clients much easier.

Supported Websocket clients:
- JavaWebSocketClient (TooTallNate/Java-WebSocket)

### DIFF
--- a/http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
+++ b/http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
@@ -1,0 +1,23 @@
+package org.http4k.websocket
+
+import org.http4k.core.Request
+import org.http4k.core.Uri
+
+typealias SymmetricWsHandler = (Request) -> Websocket
+
+fun interface SymmetricWsFilter : (SymmetricWsHandler) -> SymmetricWsHandler {
+    companion object
+}
+
+val SymmetricWsFilter.Companion.NoOp: SymmetricWsFilter get() = SymmetricWsFilter { next -> { next(it) } }
+fun SymmetricWsFilter.then(next: SymmetricWsFilter): SymmetricWsFilter = SymmetricWsFilter { this(next(it)) }
+fun SymmetricWsFilter.then(next: SymmetricWsHandler): SymmetricWsHandler = { this(next)(it) }
+
+object SymmetricWsFilters {
+    fun SetHostFrom(uri: Uri): SymmetricWsFilter = SymmetricWsFilter { next ->
+        {
+            next(it.uri(it.uri.scheme(uri.scheme).host(uri.host).port(uri.port))
+                .replaceHeader("Host", "${uri.host}${uri.port?.let { port -> ":$port" } ?: ""}"))
+        }
+    }
+}
--- a/http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
+++ b/http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
@@ -12,7 +12,7 @@ fun interface SymmetricWsFilter : (SymmetricWsHandler) -> SymmetricWsHandler {
 fun SymmetricWsFilter.then(next: SymmetricWsFilter): SymmetricWsFilter = SymmetricWsFilter { this(next(it)) }
 fun SymmetricWsFilter.then(next: SymmetricWsHandler): SymmetricWsHandler = { this(next)(it) }
 
-val SymmetricWsFilter.Companion.NoOp: SymmetricWsFilter get() = SymmetricWsFilter { next -> { next(it) } }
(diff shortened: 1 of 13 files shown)


oharaandrew314:
Tencent integration test seems to be stuck

oharaandrew314:
Is this likely to be an problem on my end?

s4nchez:
I can reproduce it locally using `./gradlew :test --tests guide.howto.testing_websocket_clients.GreetingClientTest`

I didn't have a chance to look too closely, but I can confirm that test is getting stuck, so I suspect it's either a problem in the JavaWebsocketClient or something related to the test setup.

http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
```diff
@@ -0,0 +1,22 @@
...
+typealias SymmetricWsHandler = (Request) -> Websocket
+
+fun interface SymmetricWsFilter : (SymmetricWsHandler) -> SymmetricWsHandler {
+    companion object
+}
+
+fun SymmetricWsFilter.then(next: SymmetricWsFilter): SymmetricWsFilter = SymmetricWsFilter { this(next(it)) }
+fun SymmetricWsFilter.then(next: SymmetricWsHandler): SymmetricWsHandler = { this(next)(it) }
+
+val SymmetricWsFilter.Companion.NoOp: SymmetricWsFilter get() = SymmetricWsFilter { next -> { next(it) } }
```
obecker, line 15:
```suggestion
val SymmetricWsFilter.Companion.NoOp: SymmetricWsFilter get() = SymmetricWsFilter { it }
```

http4k-realtime-core/src/main/kotlin/org/http4k/websocket/SymmetricWsHandler.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.http4k.core.Uri
+
+typealias SymmetricWsHandler = (Request) -> Websocket
+
+fun interface SymmetricWsFilter : (SymmetricWsHandler) -> SymmetricWsHandler {
+    companion object
+}
+
+fun SymmetricWsFilter.then(next: SymmetricWsFilter): SymmetricWsFilter = SymmetricWsFilter { this(next(it)) }
+fun SymmetricWsFilter.then(next: SymmetricWsHandler): SymmetricWsHandler = { this(next)(it) }
```
obecker, line 13:
```suggestion
fun SymmetricWsFilter.then(next: SymmetricWsHandler): SymmetricWsHandler = this(next)
```

oharaandrew314:
I fixed my garbage test and the build is passing now.  Thanks @s4nchez for finding it for me.

oharaandrew314:
@s4nchez Did you mean to push your server contract changes to this branch?

s4nchez:
@oharaandrew314 no, that was a mistake and I've reverted it. Sorry for the confusion.

daviddenton:
There is a question about if we can make this work more generally - ie. is the more symmetric implementation os WsHandler a better fit long term for http4k. For this, we'd also need to work out if we can mount it onto a server in the same way as the current implementation.

This could also apply to SSE as well.

oharaandrew314:
I originally wanted to make this have no breaking changes at the cost of a
more bloated API.  As it is now, I don't think it can support adding a
server.

If we're ok with breaking changes, I can come back with a suggested
solution that would replace the existing WsHandler and be truly symmetric.

daviddenton:
I think we're happy enough to support a breaking change as long as it retains the simplicity and testability. We could start with the new interfaces in the incubator package if we want to put it in incrementally and then promote it at the next major version bump. 

As long as we can retain the simplicity and testability then it's worth exploring as an option. 😄

oharaandrew314:
Ok.  I'll close this for now then.

PR #1062, opened 2024-02-21
Expose KondorJson converterFor method

tamj0rd2:
This makes it possible to use KondorJson to lookup converters for arbitrary types. This enables the possibility to generate Json Schemas for user defined classes.


FredNordin (approved):
Lgtm

PR #1063, opened 2024-02-21
Implement KondorJsonSchemaCreator

tamj0rd2:
This PR adds a KondorJsonSchemaCreator which can be used to produce well defined schemas via Kondor's inbuilt schema generation capabilities.

### DIFF
--- a/http4k-contract/openapi/src/test/resources/org/http4k/contract/openapi/v3/OpenApi3KondorTest.renders as expected.approved
+++ b/http4k-contract/openapi/src/test/resources/org/http4k/contract/openapi/v3/OpenApi3KondorTest.renders as expected.approved
@@ -1,0 +1,1112 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "title",
+    "version": "1.2",
+    "description": "module description"
+  },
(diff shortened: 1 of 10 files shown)


http4k-contract/openapi/src/test/resources/org/http4k/contract/openapi/v3/OpenApi3KondorTest.renders as expected.approved
```diff
@@ -0,0 +1,1112 @@
...
+          "200": {
+            "description": "normal",
+            "content": {
+              "application/json": {
+                "example": {
+                  "aNullField": null,
+                  "aNumberField": 123
+                },
+                "schema": {
+                  "$ref": "#/components/schemas/object-350216186"
```
tamj0rd2, line 281: When providing arbitrary json in the openapi spec instead of a data class, the generated definitionIds can change between test executions. That's because the hashCodes don't necessarily have to be consistent between execusions.

That causes this test to be flaky. I've spent quite some time looking into the issue, but I haven't been able to figure out why the Argo generated definition ids are consistent, while the ones generated via Kondor are not.

PR #1066, opened 2024-02-27
Allow adding a suffix to an approval test file name

becky-sequence:
This means that we can have different approval test files when using DynamicTest


daviddenton:
Not sure if this would be better as an extension to the approver itself instead of injecting it into each test. Can you make a case for either? 😄

becky-sequence:
> Not sure if this would be better as an extension to the approver itself instead of injecting it into each test. Can you make a case for either? 😄

@daviddenton have made some changes and added a test using `DynamicTest`. Let me know what you think!

http4k-testing/approval/src/main/kotlin/org/http4k/testing/Approver.kt
```diff
@@ -23,17 +23,24 @@ interface Approver {
...
     private val approvalSource: ApprovalSource
 ) : Approver {
 
     private val matchAllLineEndings = "\\r\\n?".toRegex()
     private fun String.normalizeLineEndings() = replace(matchAllLineEndings, "\n")
 
+    override fun withNameSuffix(suffix: String): Approver {
+        name = name + suffix
+        return this
+    }
```
flashingpumpkin, line 42: This will infinitely extend the filename name

```kotlin
approver.withNameSuffix("test1").hasApprovedContent(...) // result = "TestClass.testNametest1"
approver.withNameSuffix("test2").hasApprovedContent(...) // result = "TestClass.testNametest1test2"
```

Could we instead of returning `this`, return a a new copy of `NamedResourceApprover`? 

```kotlin
    override fun withNameSuffix(suffix: String): Approver {
        return NamedResourceApprover( name = name + suffix, ... )
    }
```

Returning a new copy means that the base name stays the same, while the suffixes change. 

The result would be the following files in your test. 

Current: 

```
ExampleApprovalTest.dynamic testA.approved
ExampleApprovalTest.dynamic testAB.approved
ExampleApprovalTest.dynamic testABC.approved
```

Fixed: 

```
ExampleApprovalTest.dynamic testA.approved
ExampleApprovalTest.dynamic testB.approved
ExampleApprovalTest.dynamic testC.approved
```

Perhaps we can also do `name = "$name.$suffix"` to separate the basename from the suffix?

flashingpumpkin (changes requested):

becky-sequence:
@daviddenton would appreciate if you could take another look at this PR please? Keen to get it merged.

daviddenton:
👍 😄

PR #1078, opened 2024-03-14
Fix bug in webdriver form submission + a method for relative Uri resolution.

gypsydave5:
Hey!

Noticed a bug in the way that the webdriver resolves a URIs from the `action` attribute of a form. Thought the best way to fix it would be to implement relative URIs as in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2), which I stuck as an extension function on a `Uri` and made public as it seemed useful.

Should have some more tomorrow.

Thanks!


daviddenton:
Looks pretty thorough. Good work! 😄

PR #1089, opened 2024-04-08
Refactoring/formatting http4k-security-oauth

dkandalov:


daviddenton:
Is this a breaking change @dkandalov ?

dkandalov:
@daviddenton it shouldn't be. I'm looking at adding PKCE https://www.oauth.com/oauth2-servers/pkce, so did some tidying first.

PR #1092, opened 2024-04-11
Change to allow for any information from FieldMetadata to appear on the final SchemaNode

BBB:
We have a requirement where we wish to append non-standard fields, and override the fields http4k attempts to derive for a given schema.

Changes to `SchemaNode` so that it is backed by a map; rather than a strict set of properties. Addition of the ability to supply an order for known properties, unknown properties appear at the bottom, sub-sorted alphabetically.

@ivanmoore @jack-bolles and myself paired on these changes, our time was donated by Ford Digital.


daviddenton:
Q. Why do we care about the sort order in the output? Seems like a lot of hassle?

PR #1094, opened 2024-04-12
OkHttp to handle NoRouteToHostException

oharaandrew314:
Since I deal with generic port-forwarded IoT devices in the hands of consumers, I'm bound to run into weird corner-cases like this one 🙄 .  Not sure if it's reproducible with other clients.


daviddenton:
👍

PR #1095, opened 2024-04-12
Fix static routes with leading slash

ArthurS1:


s4nchez:
Thank you, @ArthurS1 !

PR #1105, opened 2024-05-09
Allow overriding of `Automarshalling.asInputStream(Any)`

kwydler:
This allows some implementations to avoid creating an intermediate string when serializing directly to an `InputStream`.


daviddenton:
Thanks :)

PR #1109, opened 2024-05-20
Write CSV header even for empty list

FredNordin:
When `CsvSchema.withHeader` is used, as is for the default schema, then the resulting CSV text should include the header row even if the list of objects to write is empty.

The fix for the issue was simply to close the underlying CSV writer, which causes the headers to be written correctly.

Noticed this issue in one of the projects I'm working on. I replicated the issue using the added test, plus I also added another test to prove the defaultSchema function creates a schema with headers enabled.

Potentially breaking fix I guess.


daviddenton:
Good timing... this is a breaking release anyway!

PR #1119, opened 2024-06-12
Allows to remove all queries for a URI

dhs3000:
The `Request` already allows to remove all queries as it has the default parameter value set as "". This adds also the default to `Uri.removeQueries()` so it can remove all queries as well.

---

I'd be happy to provide a test, but I cannot find a direct test for that function to build upon, it seems to be implicitly tested through the test for `Request.removeQueries()`. Where would this new test live?


daviddenton:
There is a UriTest which can be used to host the testcase (Which is admittedly missing 🙃 )

dhs3000:
@daviddenton Thank you, I should have looked for a test by class name 🤦 I only looked for usages ;-) I added two test cases (for removing by prefix and removing all).

daviddenton:
Thank you! 🙃

PR #1125, opened 2024-06-28
WebDriver: Fix base path replacement logic for same-dir-path and dot-…

jweidler:
…path URLs

See https://datatracker.ietf.org/doc/html/rfc1808#section-4, step 6

Discovered when building a frontend that is linking exclusively via relative references between resources, for the purpose of being agnostic about the base path where it might be deployed at (reverse-proxy setups etc).


gypsydave5:
@daviddenton - this lgtm - I think the ideal would be to reuse the relative path logic I popped into `Uri`, but that can be a later refactor.

daviddenton:
👍

PR #1129, opened 2024-07-05
ApiKeySecurity that identifies a consumer and makes it available for later use

dhs3000:
The existing `ApiKeySecurity` (and filter) integrates nicely with the contract definition for the OpenAPI spec. Additionally to just blocking unauthorised requests, it would be great to identify a consumer by their API key and make them available for further processing, e.g. using it for logging.

As the two on the contract level render the same way and only the actual filter is different, I changed `ApiKeySecurity` to be an interface, so that the renderer all still work the same way. 

* I added tests for the ApiKeySecurity with consumer.
* I added tests that ensure the security renderer for the OpenAPI contract also works for this.


dhs3000:
(On first glance the failing tests seem not to be related to my changes (Playwright, Jetty)...?)

daviddenton:
Thanks for this. Will do some tweaks, but it will still be there :upsi

dhs3000:
Thanks @daviddenton :)

PR #1138, opened 2024-08-05
Add `form(vararg formData: Pair<String, String>)` request helper

tim-mortimer:
This PR adds a more compact way of adding multiple form fields to a request.

Instead of invoking `form(name: String, value: String)` per form field, a single invocation of `form(vararg formData: Pair<String, String>)` will now suffice. This has the added benefit of only parsing the form body once for all fields, as opposed to once per field.

**Considerations**

1. I've named the parameter `formData`. This aligns with how browsers summarise the collection of inputs, but the [spec](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) is a bit more vague. Should we consider changing this name?

>The application/x-www-form-urlencoded format provides a way to encode a [list](https://infra.spec.whatwg.org/#list) of [tuples](https://infra.spec.whatwg.org/#tuple), each consisting of a name and a value.

2. I've kept the signature consistent with the existing `form(name: String, value: String)` function. This is despite the underlying `Parameter` type being aliased to `Pair<String, String?>`. I think this is the right thing to do, but I thought it worth highlighting here in case it is worthy of debate.

3. I've documented usage of the new helper as part of the current how-to example. This seems consistent with the documentation of @dmcg's' `toParametersMap()` addition, but I'm open to somehow separating it out into its own example if that would be preferred.

Thanks in advance for your review of this PR!

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/core/body/FormBody.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/core/body/FormBody.kt
@@ -24,3 +24,5 @@ fun Form.toBody(): Body = Body(toUrlFormEncoded())
 fun Request.form(): Form = bodyString().toParameters()
 
 fun Request.form(name: String, value: String): Request = body(form().plus(name to value).toBody())
+
+fun Request.form(vararg formData: Pair<String, String>): Request = body(form().plus(formData).toBody())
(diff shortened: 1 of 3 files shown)


daviddenton:
It's a good point about `form(name: String, value: String)` - it should be a nullable value (as that is the underlying type) - I'll make that change later.

But anyway.. it's all good so will merge it and tweak. Welcome to the http4k contributors "hall of fame"*!

*there are no known benefits of this status 🙃

PR #1141, opened 2024-08-07
Add additional HTTP error codes

torfinnberset:
Adding additional HTTP error codes, as I encountered some of them that were not yet supported by http4k.

* [423 Locked](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/423)
* [424 Failed Dependency](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/424)
* [425 Too Early](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/425)
* [428 Precondition Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/428)
* [431 Request Header Fields Too Large](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431)


daviddenton:
👍

Issue #306, opened 2019-11-05
ServerFilters.Cors -- logic doesn't make sense


vbudilov:
Seems like the below code will always default to 'null' unless an asterisk is provided. Am I misunderstanding it? 

This code would return a null for the origins header:

```
    ServerFilters.Cors(CorsPolicy(headers = listOf("content-type", "query"),
            methods = listOf(Method.GET, Method.POST, Method.PUT, Method.DELETE, Method.OPTIONS, Method.TRACE, Method.PATCH, Method.PURGE, Method.HEAD),
            origins = listOf("localhost:8080", "mydomain.com")))
```

```
object Cors {
        private fun List<String>.joined() = joinToString(", ")

        operator fun invoke(policy: CorsPolicy) = Filter { next ->
            {
                val response = if (it.method == OPTIONS) Response(OK) else next(it)

                val origin = it.header("Origin")
                val allowedOrigin = when {
                    policy.origins.contains("*") -> "*"
                    policy.origins.contains(origin) -> origin!!
                    else -> "null"
                }

                response.with(
                    Header.required("access-control-allow-origin") of allowedOrigin,
                    Header.required("access-control-allow-headers") of policy.headers.joined(),
                    Header.required("access-control-allow-methods") of policy.methods.map { it.name }.joined()
                )
            }
        }
    }
```

daviddenton:
The `*` here takes precedence over the supplied origin *only* if the policy explicitly allows traffic from all origins. (I've also now refactored the when slightly to be a little more Kotlin-y and remove the !!.)

The docs indicate that we shouldn't be using the "null" value anyway 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

.. so maybe we should focus on if the usage here is appropriate. :)

krissrex:
> The docs indicate that we shouldn't be using the "null" value anyway https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
> 
> .. so maybe we should focus on if the usage here is appropriate and what to return if the origin does not match. Looks like it should be the list of origins that are allowed in the policy (but happy to be wrong 😄 ):

`null` seems to be a valid origin when the request is made from something that is *not* a website, such as a `file://` or `data://`. it is not the correct response for a *"CORS failed"* situation.

[...]

PR #1145, opened 2024-08-12
Include Vary header on CORS responses

ollieabbey:
As mentioned in [this issue](https://github.com/http4k/http4k/issues/306#issuecomment-1640075947).

`Vary` header should include 'Origin' in its list if a specific value is included in the `Access-Control-Allow-Origin` header, so that caches know to include the origin in the cache key. Docs [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#access-control-allow-origin) and [here](https://httpwg.org/specs/rfc9110.html#field.vary).


daviddenton:
Nice catch. Thanks! 👍

PR #1154, opened 2024-08-19
Fix #1152. Delete the temp file when the multipart form is closed ins…

### LINKED ISSUE #1152: Multipart form files call deleteOnExit() instead of being deleted when the Body is closed
This possibly caused a memory leak for long-running processes as the delete hook is stored for later processing.

daviddenton:
…tead of using deleteOnExit

### DIFF
--- a/http4k-multipart/src/main/kotlin/org/http4k/lens/parts.kt
+++ b/http4k-multipart/src/main/kotlin/org/http4k/lens/parts.kt
@@ -20,8 +20,23 @@ data class MultipartFormField(val value: String, val headers: Headers = emptyLis
     }
 }
 
-data class MultipartFormFile(val filename: String, val contentType: ContentType, val content: InputStream) : Closeable {
-    override fun close() { content.close() }
+data class MultipartFormFile internal constructor(
+    val filename: String,
+    val contentType: ContentType,
+    val content: InputStream,
+    val closeable: Closeable
+) : Closeable {
+
+    constructor(filename: String, contentType: ContentType, content: InputStream) : this(
+        filename,
+        contentType,
+        content,
+        Closeable { })
+
+    override fun close() {
+        closeable.close()
+        content.close()
+    }
 
     private data class Realised(val filename: String, val contentType: ContentType, val content: String)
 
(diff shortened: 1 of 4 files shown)


http4k-multipart/src/main/kotlin/org/http4k/multipart/DiskLocation.kt
```diff
@@ -21,9 +21,7 @@ interface DiskLocation : Closeable {
                         File.createTempFile(
                             filename ?: (UUID.randomUUID().toString() + "-"),
                             ".tmp", diskDir
-                        ).apply {
-                            deleteOnExit()
```
oharaandrew314, line 25: I think `deleteOnExit` should also be removed on line 17.

daviddenton: I think that one is ok because it applies to the temp directory as a whole (for which there is only one per JVM process)

http4k-multipart/src/main/kotlin/org/http4k/lens/parts.kt
```diff
@@ -20,8 +20,23 @@ data class MultipartFormField(val value: String, val headers: Headers = emptyLis
     }
 }
 
-data class MultipartFormFile(val filename: String, val contentType: ContentType, val content: InputStream) : Closeable {
-    override fun close() { content.close() }
+data class MultipartFormFile internal constructor(
+    val filename: String,
+    val contentType: ContentType,
+    val content: InputStream,
+    val closeable: Closeable
```
oharaandrew314, line 27: Adding the `Closeable` to the data class properties makes it so that the data class `equals` and `hashCode` will no longer work.  Evidently, there are no tests failing due to this--and they would have failed anyway because of the `InputStream`--but it makes me question why this was a `data class` to being with.

http4k-multipart/src/main/kotlin/org/http4k/multipart/DiskLocation.kt
oharaandrew314: `deleteOnExit` is being called on `Files.createTempDirectory("http4k-mp")`, which creates a new temp directory in `java.io.tempdir` for every invocation, each with a unique suffix.

PR #1159, opened 2024-08-20
Share DiskLocation across all Multipart lens extractions

oharaandrew314:
- Multipart Lenses will now share a single `DiskLocation` for all lens extractions
- Fixed an order-of-operations error while deleting multipart files
- Ensure disk-backed form fields are deleted correctly

### DIFF
--- a/http4k-multipart/src/main/kotlin/org/http4k/core/MultipartFormBody.kt
+++ b/http4k-multipart/src/main/kotlin/org/http4k/core/MultipartFormBody.kt
@@ -19,19 +19,14 @@ sealed class MultipartEntity : Closeable {
     abstract val name: String
     internal abstract fun applyTo(builder: MultipartFormBuilder): MultipartFormBuilder
 
-    data class Field(override val name: String, val value: String, val headers: Headers = emptyList()) :
-        MultipartEntity() {
-        override fun close() = Unit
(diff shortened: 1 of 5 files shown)


oharaandrew314:
If I'm reading the build failure correctly, it might have just been a transient netty error.

oharaandrew314:
Now it's tencent failing?

daviddenton:
I've fixed the test, but I don't think this will work right now because if 2 requests come in with the same filenames, they will share each others files. The File itself is exposed through the API so we can't just randomise the name unfortunately :(

oharaandrew314:
So I guess we could reject this and revert your previous update, but that still leaves us where we were before.  I wonder if it's worth undoing just the directory change, and seeing if the other updates make a difference.

daviddenton:
try it and see! 🙃

oharaandrew314:
Which test was actually failing?  For some reason, it wasn't obvious to me in the CI logs, and several multipart tests always fail for me on windows anyway.

daviddenton:
There were 2. One of the failing runs had them if you look (am AFK)

oharaandrew314:
Good to merge now.

daviddenton:
👍

daviddenton:
thanks for sticking with it!

oharaandrew314:
If I don't, then I have to keep manually recovering containers!

PR #1172, opened 2024-09-07
Create unified Websocket factory interface

oharaandrew314:
This will make it easier to pass websocket clients around; a stepping stone towards a symmetric `WsHandler`.


daviddenton:
the failure here is because of a timeout in jetty11, which is really slow to shutdown.

PR #1181, opened 2024-09-30
Gzip mixed compression mode

IlyaNerd:
I'm ok to get a push back on this, but we discovered for our streaming functionality, that gzip filter by default buffers everything in memory. We can just pass streaming for all routes... But we decided to implement a similar functionality in our code - to decide the mode for each request based on the body given.

Also it feels like this can be the default option.

### DIFF
--- a/http4k-core/src/main/kotlin/org/http4k/filter/Gzip.kt
+++ b/http4k-core/src/main/kotlin/org/http4k/filter/Gzip.kt
@@ -3,6 +3,7 @@ package org.http4k.filter
 import org.http4k.core.Body
 import org.http4k.core.Request
 import org.http4k.core.Response
+import org.http4k.core.StreamBody
 import java.io.ByteArrayInputStream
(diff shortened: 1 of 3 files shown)


http4k-core/src/main/kotlin/org/http4k/filter/Gzip.kt
```diff
@@ -16,6 +17,18 @@ import java.util.zip.Deflater.DEFAULT_COMPRESSION
 import java.util.zip.GZIPInputStream
 import java.util.zip.GZIPOutputStream
 
+/**
+ * GzipCompressionMode represents a choice between working lazily with streams when gzipping or eagerly storing the body contents in memory.
```
IlyaNerd, line 21: This is mostly a copy paste from BodyMode

daviddenton:
Nope - this looks good! Thanks!

We might look at making it the default mode in future.

😄

PR #1186, opened 2024-10-08
Better diff messages on Approver missing/mismatch

FredNordin:
Previously, when there was a mismatch in the Approver, the difference could be seen in the console, but because of the message format and exception type, IDEs like IntelliJ would not be able to offer a diff view (by either clicking on the message in the console or typing Cmd+D when selecting the test in the test runner).

With the changes provided in this PR, that is now possible. For example, if the approved file does not exist, it looks like:
image

A mismatch looks like:
image

The change also removes the need for the ApprovalFailed exception class.

It's just a minor change, but one that really makes a difference when using approval test, as one can simply Cmd+D on the failing test method, instead of having to navigate to the actual/approved files in the navigator and running a diff on the files.


daviddenton:
Looks great! thanks 😄

PR #1196, opened 2024-10-24
Add resilience4j timeout filter

ollieabbey:
Allows using resilience4j timeouts through a simple filter.

Standard usage:

```kotlin
ResilienceFilters.TimeLimit(TimeLimiter.of(Duration.ofMillis(500)))
  .then {
    val upstreamResponse = callSomeSlowApi()
    Response(OK).body(upstreamResponse)
  }
```

I don't know if allowing passing in the executor service is overkill, but it allows users a bit more control over what threads are used if they want it. If you'd rather not use that, just remove the second & third commits.

If there's any other reason this functionality wasn't already included or you just don't want it, feel free to close this PR 🙂


daviddenton:
No reason it wasn't included. Thanks! :)

PR #1219, opened 2024-11-27
Utility function to get the structure suffix from a content type

ollieabbey:
Named as per the [media type spec](https://www.rfc-editor.org/rfc/rfc6838#section-4.2.8).

Lets you write stuff like 

```kotlin
if (response.contentType()?.structure() == "json") {
  // do something
}
```


daviddenton:
👍

PR #1223, opened 2024-11-27
Fix #1221 Lack of ability to automatically generate valid Open API schemas for downloading files

### LINKED ISSUE #1221: Lack of ability to automatically generate valid Open API 3.0 schema for downloading files.
Lack of ability to automatically generate valid Open API 3.0 schema for downloading files.

For example when I have an endpoint returning content type `application/pdf ` (or `image/png`, or ...) and I expect:
```json
{
  "/pdf": {
    "get": {
      "summary": "Get PDF file",
      "parameters": [],
      "responses": {
        "200": {
          "description": "PDF file",
          "content": {
            "application/pdf": {
              "schema": {
                "type": "string",
                "format": "binary"
              }
            }
          }
        },
        "404": {
          "description": "File not found"
        }
      }
    }
  }
}
```

Currently if I try the following:
```
  returning(
      status = OK,
      body = Body.binary(APPLICATION_PDF).toLens() to "pdf".toByteArray().inputStream()
  )
```

I get exception `Caused by: com.github.underscore.Json$ParseException: Unexpected end of input at 1:-1`

hektorKS:
Issue link: https://github.com/http4k/http4k/issues/1221

This change we will also allow to use [openapi-generator](https://github.com/OpenAPITools/openapi-generator) to generate file types e.g. Blob for Typescript.

### DIFF
--- a/core/contract/src/main/kotlin/org/http4k/contract/openapi/v3/OpenApi3.kt
+++ b/core/contract/src/main/kotlin/org/http4k/contract/openapi/v3/OpenApi3.kt
@@ -47,6 +47,7 @@ import org.http4k.lens.ParamMeta.EnumParam
 import org.http4k.lens.ParamMeta.ObjectParam
 import org.http4k.lens.ParamMeta.StringParam
 import org.http4k.lens.WebForm
+import java.io.InputStream
 import java.math.BigDecimal
 import java.math.BigInteger
 import java.util.Locale
@@ -189,10 +190,10 @@ class OpenApi3<NODE : Any>(
         }.toMap()
 
     private fun List<HttpMessageMeta<Response>>.collectSchemas() = groupBy { CONTENT_TYPE(it.message) }
-        .mapValues {
-            when (it.value.size) {
-                1 -> it.value.first().toSchemaContent()
-                else -> OneOfSchemaContent<NODE>(it.value.map { it.toSchemaContent() })
+        .mapValues { (_, metas) ->
+            when (metas.size) {
+                1 -> metas.first().toSchemaContent()
+                else -> OneOfSchemaContent<NODE>(metas.map { meta -> meta.toSchemaContent() })
             }
         }
         .mapNotNull { i -> i.key?.let { it.value to i.value } }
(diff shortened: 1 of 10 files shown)


core/contract/src/main/kotlin/org/http4k/contract/openapi/v3/OpenApi3.kt
```diff
@@ -315,18 +316,28 @@ class OpenApi3<NODE : Any>(
...
-            ?: message.bodyString().toSchema(definitionId)
+        fun exampleSchemaIsValid(schema: JsonSchema<NODE>) = when (example) {
+            is Array<*>, is Iterable<*> -> !json.fields(schema.node).toMap().containsKey("\$ref")
+            else -> apiRenderer.toSchema(object {}, refModelNamePrefix = schemaPrefix) != schema
+        }
 
-        return SchemaContent(jsonSchema, message.bodyString().safeParse())
+        return when (example) {
+            is InputStream -> NoSchema(
+                schema = json {
```
daviddenton, line 326: Is there a type we can use instead of manually creating the JSON here?

daviddenton:
Well done on taking this on - it's not exactly simple code!

Just one comment from me, but other than that it looks 👍

core/contract/src/main/kotlin/org/http4k/contract/openapi/v3/OpenApi3.kt
hektorKS: Hmm, I used the same approach as in other places in this file:
e.g. 
`val noSchema = consumes.map { it.value to NoSchema(json { obj("type" to string(StringParam.value)) }) }`

hektorKS: There is space to refactor this code for sure, but probably we prefer small MRs ;)

daviddenton: ok - I was just checking and not wanting to go and look!

PR #1231, opened 2024-12-04
Updates to formatting of cookies

ollieabbey:
1. When the cookie has no attributes, don't end the string in a trailing `; `. 
	- As per [the spec](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1), specifically the line `set-cookie-string = cookie-pair *( ";" SP cookie-av )`
	- I guess there's a chance this may break tests for consumers that are currently asserting the precise cookie string value?
2. Allow the setting of an unquoted cookie on a `Response`
3. Allow the invalidation of a cookie with a path

### DIFF
--- a/core/core/src/main/kotlin/org/http4k/core/cookie/Cookie.kt
+++ b/core/core/src/main/kotlin/org/http4k/core/cookie/Cookie.kt
@@ -41,7 +41,7 @@ data class Cookie(
         appendIfTrue(secure, "secure")
         appendIfTrue(httpOnly, "HttpOnly")
         appendIfPresent(sameSite, "SameSite=$sameSite")
-    }.joinToString("; ")
+    }.joinToString("") { "; $it" }
 
     companion object {
         fun parse(cookieValue: String): Cookie? {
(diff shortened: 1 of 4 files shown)


ollieabbey:
Looks like the build has failed from a timeout, but I think there's another failing test in there I'll push a fix for.

daviddenton:
> Looks like the build has failed from a timeout, but I think there's another failing test in there I'll push a fix for.

yes - we're having problems with builds right now - we're looking into it. don't worry about the timeout and just retry it if it fails.

ollieabbey:
The test job now isn't timing out, but it's failing on a test that passes locally. Unfortunately I think it's having an issue publishing the test report.

> 🚀 Publish results
  ℹ️ - JUnit Test Report - 4767 tests run, 4406 passed, 360 skipped, 1 failed.
     🧪 - JettySseTest | java.lang.AssertionError: expected: a value that not contains substring "exception"
  ℹ️ - JUnit Test Report - Updating checks (Annotations: 1)
  Error: ❌ Failed to create checks using the provided token. (HttpError: Resource not accessible by integration - https://docs.github.com/rest/checks/runs#update-a-check-run)
  Warning: ⚠️ This usually indicates insufficient permissions. More details: https://github.com/mikepenz/action-junit-report/issues/23

I'm also struggling to rerun the job - I'm not sure I've got permissions to rerun?

ollieabbey:
Hiya @daviddenton, looks like this is passing the build now whenever you're free to take a look. Any issues give me a shout and I can take a look, or amend/close this wherever you please 🙂

s4nchez:
I've checked too and the changes are good. Thank you, @ollieabbey :)

PR #1241, opened 2024-12-15
PKCE-based Authorization for AWS SSO

aharin:
Adds support for PKCE-based Authorization for AWS SSO. This pull request also adds support for caching of AWS SSO credentials in ~/.aws/sso/cache and adds a login disabled mode, where provider will fail if application has not been previously logged in into AWS. This works in line with AWS command line tools.


daviddenton:
Thanks for the contribution and sorry it's taken a bit to get to it, but it's quite involved so we wanted to make sure it was all good. 🙃

PR #1262, opened 2025-01-17
Ensure module names are correctly generated on Windows machines

kwydler:
Encountering the following error when building the project on Windows.

```
A problem occurred configuring project ':http4k-connect\ai\anthropic\client'.
> The project name 'http4k-connect\ai\anthropic\client' must not contain any of the following characters: [/, \, :, <, >, ", ?, *, |]. Set the 'rootProject.name' or adjust the 'include' statement (see https://docs.gradle.org/8.11.1/dsl/org.gradle.api.initialization.Settings.html#org.gradle.api.initialization.Settings:include(java.lang.String[]) for more details).
```

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -28,6 +28,7 @@ rootDir.walkTopDown()
     .filterNot { dir -> dir == rootDir || exclusions.any { dir.absolutePath.contains(it) } }
     .forEach {
         val moduleName = it.relativeTo(rootDir).path
+            .replace(File.separatorChar, '/')
             .removePrefix("core/") // remove core ecosystem modules as they have no prefix to add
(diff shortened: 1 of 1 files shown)


s4nchez:
Thank you for the PR, @kwydler. 

Please notice we're not actively supporting Windows as the http4k core team is mostly on Mac/Linux, but we're always happy to incorporate changes to make life easier for users/contributors.

daviddenton:
I think we might need to change all of the references in the gradle file to use the file separator instead of the '/'. Can you confirm if that's the case @kw

kwydler:
I checked most of the Gradle files and I think everything else is fine. Anything that is using the Gradle API or Java `File` API seems to be fine with the `/` separator even on Windows. It does look like some tests are failing on Windows. I'll take a look at those and create another PR if needed.

PR #1266, opened 2025-01-20
Passthrough calls to `close` and `flush` to ensure output stream is p…

kwydler:
…roperly closed

This caused the `File.delete()` call in certain `FileReadWriteResourceTest` tests to not remove the temporary file as expected, resulting in test failures during subsequent runs due to the existing file.


daviddenton:
👍

PR #1267, opened 2025-01-21
Added lens and support for serialisation of java.time.Period

aharin:
This pull request adds lenses and support for serialisation of java.time.Period.


daviddenton:
👍

PR #1280, opened 2025-01-29
Fix mismatched request signature error when s3 path contains plus symbol

aharin:
At the moment amazon s3 client cannot handle files that contain '+' character in the file name. Request fails with SignatureDoesNotMatch error. There seem to be a known issue with AWS not handling + sign in request path in a standard way (https://jamesd3142.wordpress.com/2018/02/28/amazon-s3-and-the-plus-symbol/). This issue is also discussed here:  https://github.com/akka/alpakka/issues/838 . The solution is to replace '+' with '%2B' after request has been signed.


s4nchez:
I have confirmed the issue and tested the fix. Thank you for spotting and fixing it, @aharin!

PR #1297, opened 2025-02-21
basic support for ses v2 sendEmail

### LINKED ISSUE #1203: Support V2 of the SES API (Fake and client)
Actions to be implemented:

- SendEmail

[API v2](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_Operations.html)

oharaandrew314:
Add support for SES V2 `sendEmail`.  The fake does not yet support templated emails.

Closes #1203


connect/amazon/ses/client/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 
 dependencies {
     api(project(":http4k-connect-amazon-core"))
+    implementation("com.sun.mail:javax.mail:_")
```
oharaandrew314, line 13: Optional.  Includes helper to generate a `RawMessagebase64` from a `javax.mail.Message`.

oharaandrew314:
I don't know why my `RealSESTest` is running.  What did I miss?

connect/amazon/ses/client/build.gradle.kts
aharin: Wasn't javax namespace deprecated? I think com.sun.mail:jakarta.mail is now preferred.

oharaandrew314: @aharin Good point.  I've also added optional support for jakarta.  Jakarta mail doesn't like when javax mail is on the same classpath, so javax mail is now added as `compileOnly`, and can't have test coverage.

daviddenton: Q - why would we use javax mail at all if it's deprecated now? Shouldn't we just use jakarta as that's the future? :)

oharaandrew314: For the same reason we still support javax servlets? :shrug: I'll concede it's much easier to upgrade to jakarta mail than to jakarta servlets though.  It's just an optional dependency for a small convenience function, but I don't mind removing it, if you want.

daviddenton: Nah - the only reason we've still got javax (which arguably should have been removed as a part of v6) was for backwards compatibility.,This is new, so I say we just ditch it.

oharaandrew314: Ok, done

oharaandrew314:
I also found the `Base64Blob` I lost track of, and replaced the `RawMessageBase64` with it.  Not planning any more updates after this.

connect/amazon/ses/client/build.gradle.kts
daviddenton: 👍

PR #1299, opened 2025-02-24
Add `lenient()` for adapter for Lenient strictness in ConfigurableMoshi

overfullstack:
This change should help in parsing JSON in lenient mode (as stated here in Moshi: [JsonReader.kt](https://github.com/square/moshi/blob/256b13b3bfd6cb5557c857496e674b0b1ab98d10/moshi/src/main/java/com/squareup/moshi/JsonReader.kt#L200-L200)), as per the lenient param in the ConfigurableMoshi constructor.


daviddenton:
We can see what you've done form the commit log. What's more interesting is why it was done, and what improvement does it make? 🙃

overfullstack:
Sorry @daviddenton, I assumed it's implicit :). 
I expected lenient param in the constructor to help me parse JSON in a lenient format, as stated here:
[JsonReader.kt](https://github.com/square/moshi/blob/256b13b3bfd6cb5557c857496e674b0b1ab98d10/moshi/src/main/java/com/squareup/moshi/JsonReader.kt#L200-L200)

I currently have a use-case where my JSON contains End of line comments, which I wish to parse in Lenient mode

PR #1300, opened 2025-02-25
Add call to `withInstanceMetadataServiceMappings()` when constructing…

kwydler:
… `InstanceMetadataServiceMoshi`

Without this call the `GetInstanceIdentityDocument` action throws an exception when deserializing `ImageId`.


daviddenton:
👍

PR #1302, opened 2025-02-26
Added support for SystemManager.getParameters endpoint

aharin:
At the moment only getParameter that can retrieve a single Parameter is supported. This pull request adds support for getParameters endpoint that can retrieve multiple parameters at one go. It is also useful for retrieving a parameter that might not exist as it returns collection of parameters and does not throw if requested parameter does not exist.


daviddenton:
👍

PR #1308, opened 2025-03-01
OpenApi3 Renderer test for Moshi

potfur:
Test case for OpenApi V3 renderer for Moshi


daviddenton:
I'm finally gonna pick this up as I think it's time to maybe think about a post-Jackson era :)

PR #1318, opened 2025-03-18
Document helidon websocket server bug

oharaandrew314:


oharaandrew314:
Closing since #1321 fixes the bug

PR #1320, opened 2025-03-19
Some additions to http4k-connect-storage

ollieabbey:
- Allow passing a custom duration function when constructing a Redis instance, so you can cache an object for a different duration depending on its properties.
- In org.http4k.traffic, add an extension to transform a Storage instance into a ReadWriteCache. http4k-connect-storage-core is included as a compileOnly dependency so you don't automatically ship these two modules together, but if you've got them both you can use this function

These two additions between them mean you can do things like easily use REDIS as a cache in front of an HTTP client, caching the responses for different durations depending on their max-age.


daviddenton:
👍

PR #1321, opened 2025-03-19
Helidon Websocket Server to handle headers properly

oharaandrew314:
Helidon does provide the headers in an `onHttpUpgrade` hook.  However, when the `HelidonToHttp4kWebSocketListener` is shared for an entire server, it's (likely) impossible to guarantee we can match upgrade requests to their sessions.

To get around that, the `HelidonToHttp4kWebSocketListener` is now unique to each websocket.  As a result, we no longer need to reflectively access internals to get the upgrade request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ changes with their rationale when appropriate:
 - **http4k-connect-openai-*** : [Breaking] Rename `FinishReason` to `StopReason`
 - **http4k-connect-ai-anthropic-*** : [Breaking] Replaced `Schema` with a simple `Map<String, Any>`, as we are unifying the types across all AI modules. A new Schema type will be introduced in a future release. Replace existing Schema usage with a Map<String, Any> for now.
 - **http4k-connect-ai-anthropic-*** : [Fix] Correct name of input_schema field in Tool.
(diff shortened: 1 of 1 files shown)


oharaandrew314:
I don't think the netty failure is related.

core/bridge/helidon/src/main/kotlin/org/http4k/bridge/HelidonToHttp4kWebSocketListener.kt
```diff
@@ -1,71 +1,61 @@
...
-    private val prologue = WsConnection::class.java.getDeclaredField("prologue").also {
-        it.isAccessible = true
-    }
-    private val uriPath = HttpPrologue::class.java.getDeclaredField("uriPath").also {
-        it.isAccessible = true
-    }
-    private val rawQuery = HttpPrologue::class.java.getDeclaredField("rawQuery").also {
-        it.isAccessible = true
+    override fun onHttpUpgrade(prologue: HttpPrologue, headers: Headers): Optional<Headers> {
+        assert(!complete) { "Cannot be reused for multiple sessions" }
```
oharaandrew314, line 56: I could use an `AtomicBoolean` to make this atomic, but given that it's a testing safeguard, I don't think the performance impact is worth it.

daviddenton:
👍

PR #1336, opened 2025-04-09
Correct SQS SendMessage request DTO

mghwang:
An InvalidParameterException occurs due to the invalid SQS SendMessage request DTO field name 'MessageDeDuplicationId' when we try to send a message to a queue without ContentBasedDeduplication feature.


mghwang:
The actual error message is as follows:

> RemoteFailure(method=POST, uri=, status=400 , message={"__type":"com.amazon.coral.service#InvalidParameterValueException","message":"The queue should either have ContentBasedDeduplication enabled or MessageDeduplicationId provided explicitly"})

daviddenton:
Thank. Until we release this, suggest that you can use a dupilcated action class with the corrected name. :)

PR #1345, opened 2025-04-22
Allow the `modelNamer` to be configured when using `jacksonExt.OpenApi3`

mbcltd:


daviddenton:
It was in conflict so I've reapplied the change manually. :)

Issue #943, opened 2023-07-05
Http4k OpenTelemetry Metrics deviate from OTel Semantic Conventions


krissrex:
## Background

OpenTelemetry has a recommended way of naming metrics: the [Semantic Conventions](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/http-metrics).

They also have a shared implementation of metric factories used by their intrumentors:  [client](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java) and  [server](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java) metric factories in `open-telemetry/opentelemetry-java-instrumentation` `instrumentation-api-semconv` module.
I don't believe it is published as a dependency we can use, but the settings used could be copied.

## Problem

Http4k defines `http.server.request.latency`, where semconv says [`http.server.request.duration`](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/http-metrics/#metric-httpserverduration).

https://github.com/http4k/http4k/blob/4185ce559556ca4583aefbb78bed590c207fa017/http4k-core/src/main/kotlin/org/http4k/metrics/MetricsDefaults.kt#L16-L31

This is defined in the `http4k-core`, then used in `http4k-opentelemetry`:
https://github.com/http4k/http4k/blob/4185ce559556ca4583aefbb78bed590c207fa017/http4k-opentelemetry/src/main/kotlin/org/http4k/filter/metricsFiltersOpenTelemetryExtensions.kt#L42-L48

### Rename
I doubt we can use the same implementation as `instrumentation-api-semconv`.
Instead, i suggest renaming the metric. 

> Note that the unit recommended is `s`, but the reference `instrumentation-api-semconv` may use `ms` for backwards-compatible reasons.

### Explicit buckets
I suggest we use the defined histogram buckets from SemConv: `[ 0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ]`.

Http4k would do it here:
[...]

PR #1347, opened 2025-04-29
implement OpenTelemetry Semantic Convention compatible required server and client metric

dzappold:
related to #943


dzappold:
After some by far to complex approaches I came up with this filter version to support OpenTelemetry Semantic Convention compatible metric. 
Neither tracing nor Micrometer metrics are touched so far.

daviddenton:
Seems wrong to have the otel dependencies present in the ops/core module - was that intended?

dzappold:
You are right, idea behind that is to use it later for the micrometer part as well. 
Which option would have prefer?
- keeping the metrics definitions inside of opentelemetry itself 
- having a separate module for metrics definitions to use them later in micrometer module
- something else

daviddenton:
For the moment, let's keep everything inside  OTL and we can iterate :)

> You are right, idea behind that is to use it later for the micrometer part as well. Which option would have prefer?
> 
> * keeping the metrics definitions inside of opentelemetry itself
> * having a separate module for metrics definitions to use them later in micrometer module
> * something else

PR #1348, opened 2025-04-29
Add support for root-level GraphQL extensions field in GraphQLResponse

joostbaas:
The [spec](https://spec.graphql.org/October2021/#sec-Response-Format) says: The response map may also contain an entry with key extensions. This entry, if set, must have a map as its value.

This PR will copy the extensions from the ExecutionResult into the GraphQLResponse.

For instance the [Apollo-tracing](https://github.com/apollographql/apollo-tracing?tab=readme-ov-file) compatible [TracingInstrumentation class from graphql-java](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/instrumentation/tracing/TracingInstrumentation.java) uses this field, which can then be displayed by the already included graphql-playground UI.

### DIFF
--- a/core/api/graphql/src/main/kotlin/org/http4k/graphql/model.kt
+++ b/core/api/graphql/src/main/kotlin/org/http4k/graphql/model.kt
@@ -21,7 +21,10 @@ data class GraphQLRequest(
     }
 }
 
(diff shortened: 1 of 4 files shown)


daviddenton:
👍

PR #1404, opened 2025-08-27
Lenses and content negotiation logic for accept-like headers

npryce:
Lenses for accept-like headers, and a model for "simple" ranges suitable for  Accept-Charset, Accept-Encoding, and Accept-Language with the Basic Filtering content negotiation scheme. Functions that interpret priority lists, suitable for implementing content negotiation in the application layer.


npryce:
Closing while I work on negotiated routing, which will probably change the API a bit

PR #1410, opened 2025-09-05
Add support for HtmlFlow templates

BernardoPe:
Adds HtmlFlow Templates support, as discussed [here](https://github.com/xmlet/HtmlFlow/issues/122#issuecomment-3185468741)

Supported methods:
- `CachingClasspath`: Scans package paths for `HtmlView` definitions with view caching
- `HotReloadClasspath`: Same scanning with hot reload capability  
- `Caching`: Converts directory paths to package paths for classpath-based scanning
- `HotReload`: Same as `Caching` with hot reload enabled
- `HtmlView<T>.renderer()`: Creates a `TemplateRenderer` directly from an `HtmlView`

### DIFF
--- a/core/template/htmlflow/build.gradle.kts
+++ b/core/template/htmlflow/build.gradle.kts
@@ -1,0 +1,13 @@
+description = "http4k HtmlFlow templating support"
+
+plugins {
+    id("org.http4k.community")
+}
+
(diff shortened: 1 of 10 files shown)


daviddenton:
👍

PR #1415, opened 2025-09-09
Handle additional connection issues in client HttpHandlers

### LINKED ISSUE #1414: Clients throw exceptions for HTTP connection issues
Some of the client `HttpHandler` throw an exception on bad responses from the server, e.g. connection reset.

Clients affected:
* java
* java8
* helidon
* jetty
* okhttp

It can be reproduced with WireMock:

```kotlin
@Test
fun `connection reset throws exception`() {
    val server = WireMockServer(WireMockConfiguration.options().dynamicPort())
    server.start()

    try {
        server.stubFor(
            get(urlEqualTo("/connectionIssue"))
                .willReturn(
                    aResponse()
                        .withFault(Fault.CONNECTION_RESET_BY_PEER)
                )
        )
        client(Request(Method.GET, "http://localhost:${server.port()}/connectionIssue"))
    } finally {
        server.stop()
    }
}
```

alex859:
Fixes #1414 

I used WireMock to reproduce the issues, it was quite handy, though I’m happy to explore alternatives if you prefer.

I haven’t added assertions on the status description, since different clients throw different exceptions with varying messages.

One concern I have is that catching those exceptions hides the root cause, which can be useful when investigating issues.

Looking forward to your feedback. Thanks! ☺️

### DIFF
--- a/core/client/apache/src/main/kotlin/org/http4k/client/ApacheClient.kt
+++ b/core/client/apache/src/main/kotlin/org/http4k/client/ApacheClient.kt
@@ -15,6 +15,7 @@ import org.apache.hc.client5.http.impl.classic.HttpClients
 import org.apache.hc.core5.http.ClassicHttpResponse
 import org.apache.hc.core5.http.Header
 import org.apache.hc.core5.http.HttpResponse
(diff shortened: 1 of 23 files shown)


daviddenton:

PR #1430, opened 2025-10-28
AWS Xray requires message.type and messaging.message_payload_size_byt…

aharin:
Currently response content length is not reported in AWS Xray. According to https://aws-otel.github.io/docs/getting-started/x-ray: 

> If message.type attribute key exists and value = "RECEIVED" then http.content_length is set to messaging.message_payload_size_bytes if it is set; otherwise http.content_length is set to 0. 

This pull request sets message.type and messaging.message_payload_size_bytes attributes which fixes reporting of response content length in AWS Xray.


daviddenton:
👍👍

Issue #1426, opened 2025-10-25
Avoiding repeated copies of `HttpMessage.bodyString()`


hermannm:
Hi,

First off: Great work on the library, I love using http4k!

In the http4k applications that I work on, we have filters that wish to inspect request/response bodies (for logging, among other things). These typically want to work with bodies as strings, and therefore call `HttpMessage.bodyString()`. The [current implementation of `HttpMessage.bodyString()`](https://github.com/http4k/http4k/blob/6.20.0.3/core/core/src/main/kotlin/org/http4k/core/http.kt#L170) calls `String(body.payload.array())`, which copies the body into a new string. When you call this method at multiple places in your filter stack, in addition to your main HTTP handler, multiple copies of the body will be made. This can increase memory pressure when HTTP bodies are a significant part of your application's memory churn.

To alleviate this issue, we want to save the result of `HttpMessage.bodyString()` the first time it's called, so that subsequent calls will not copy the body again. We've found a couple ways to achieve this with http4k's current interfaces:

- Using a `RequestKey`/`ResponseKey`: This allows us to attach the body string to the request/response, and extract it   later when retrieving the body string again.
  - One downside of this approach is that `HttpMessage.bodyString()` will still copy the body, so we have to use a custom function in order to get the body string from our `RequestKey`/`ResponseKey`. Enforcing this across all our filters and applications can be difficult.
- Wrapping `Request`/`Response`: This solves the downside of the previous approach, since we can override `HttpMessage.bodyString()` to get a cached result if it's already been called.
  - But a downside of this approach is that we must be careful about how we wrap `Request`/`Response`. For example, http4k has its own wrappers, `RequestWithContext`/`ResponseWithContext`, that must always be the "outer" wrapper, since http4k [does checks like `request is RequestWithContext`](https://github.com/http4k/http4k/blob/6.20.0.3/core/core/src/main/kotlin/org/http4k/lens/request.kt#L22), which will fail if the `RequestWithContext` itself is wrapped.

As you can see, neither of these solutions are ideal. One of the issues here is that `bodyString()` works on the `HttpMessage` level - `Body` doesn't have a say in how it's converted to a string, since `HttpMessage` does the `String` construction. If `HttpMessage` instead delegated that responsibility to the `Body` interface, then it would be easier to provide a `Body` implementation that saves the first result of calling `bodyString()`, instead of wrapping the full `Request`/`Response`. Therefore, I propose the following:

[...]

PR #1436, opened 2025-11-03
Add `Body.text` property for getting the string representation of a body

hermannm:
- Add `text` property to the `Body` interface, with a default implementation for backwards compatibility
- Change `HttpMessage.bodyString()` to call `body.text`
- Override `text` in `MemoryBody` and `StreamBody` to memoize the first result of calling it, to avoid the body being copied on each call

This resolves issue #1426.

`BodyTest` has been expanded to cover the new API surface, and test that memoization works as expected.

I've tried to keep the code style of my implementation the same as what I've seen elsewhere in http4k, but let me know if you'd like me to do things differently.

I made `text` a property instead of a method, because you seemed to want that [in your comment on my issue](https://github.com/http4k/http4k/issues/1426#issuecomment-3473911653).


daviddenton:

PR #1451, opened 2025-12-05
Allow consistentRead in table mapper for get operations

PaulWMR:
While the DynamoDbTableMapper provides a very nice, simple API, adding the consistentRead parameter to allow strongly consistent reads can be very useful. It would be very nice not to have to go back to the the regular DynamoDb client for that.

If you feel this opens a can or worms when it comes to the simplified uses of the DynamoDbTableMapper and should stick the the simplest of operations, feel free to let me know and close the PR. I do believe this addition should keep the table mapper simple enough, which is why I'm raising this PR, but I can see how opinions could vary.


daviddenton:
I think this is fine - it's backwardly compatible and not too unwieldy :) Thanks!

daviddenton:

PR #1467, opened 2026-01-13
feat(datastar): only include useViewTransition and mode fields in events if values are overridden

Christian-Broomfield:
## Description
In a recent conversation with the author of Datastar, it was revealed that sending the default `useViewTransition false` line in a datastar event caused the view transition to be enabled. [A bug report has been created here.](https://github.com/starfederation/datastar/issues/1120) Nevertheless, it's still desirable to avoid sending `useViewTransition` and `mode` if the defaults aren't overridden. 

This PR changes `DatastarEvent` so that these values are only sent if they're not `mode outer` or `useViewTransition false`.


daviddenton:
we were having some problems with our build, so I merged this, but it looks like all of the Datastar contract tests were actually failing so I've had to revert the commit. If you can work out and fix the problem then we can take another run at it 🙃

daviddenton:
I'd initially reverted this but the fix was simple so it's now in master. Thanks for the contrib.

daviddenton:

PR #1491, opened 2026-02-22
Default to application/octet-stream when multipart file part has no Content-Type

### LINKED ISSUE #1486: MultipartFormBody - Files Missing `contentType` Stop Parsing
I've had trouble handling multipart form submissions from different tools, curl, browsers, and Deno work well, but Httpie appears to stop multipart streaming when attempting to upload a binary file lacking in type info:

```sh
# Works
curl \
  -u user:password \
  -F "hellp=world" \
  -F "file@blob.elf" \
  https://development.test.server

# Does not work
http -a user:password \
  -f POST https://development.test.server \
  hello=world \
  file@blob.elf
```

As part of a very basic web server, requests that are posted to a specific route with an appropriate "multipart/form-data` content type run through this iterator, accumulating fields and files:

```kotlin
request.multipartIterator().asSequence().forEach {
    when (it) {
        is MultipartEntity.Field -> {
            // Do something
        }
        is MultipartEntity.File -> {
            // Do other thing
        }
    }
}
```

Digging deeper into the `multipartIterator`, I suspect there may be a risky non-null assertion when creating a `MultipartFormFile` - the incoming `StreamingPart` may have a null `contentType` set on it. Should this be defaulted to something like "application/octet-stream"?

Testing this theory, I tried the Httpie request again, but qualified the file upload with a type:

```sh
# Works
http -a user:password \
  -f POST https://development.test.server \
  hello=world \
  file@'blob.elf;type=application/octet-stream'
```

We get all sorts of requests from tools of varying quality, hopefully the core multipart area can be made more defensive against missing type info on files. Happy to answer any further questions or provide examples.

andy-barlow:
## Summary

- When a multipart file part has no `Content-Type` header, the parser now defaults to `application/octet-stream` instead of throwing a NullPointerException
- Fix applied to both the streaming (`multipartIterator()`) and buffered (`MultipartFormBody.from()`) parsing paths
- Added tests for both code paths

Fixes #1486

### DIFF
--- a/core/core/src/main/kotlin/org/http4k/lens/header.kt
+++ b/core/core/src/main/kotlin/org/http4k/lens/header.kt
@@ -86,7 +86,7 @@ object Header : BiDiLensSpec<HttpMessage, String>(
         with(it.split(";").mapNotNull { it.trim().takeIf(String::isNotEmpty) }) {
             first() to drop(1).map {
                 with(it.split("=")) {
-                    first() to if (size == 1) null else drop(1).joinToString("=")
+                    first() to if (size == 1) null else drop(1).joinToString("=").removeSurrounding("\"")
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


daviddenton:

PR #1518, opened 2026-04-19
OpenAI and vLLM spec additions

torfinnberset:
OpenAI spec additions:
- Request: logprobs, top_logprobs fields
- Response: ChoiceLogProbs/TokenLogProb/TopLogProb on Choice
- Response: reasoning field on ChoiceDetail (OpenAI GPT-OSS)

Fix ResponseFormat.JsonSchema to use typed JsonSchemaSpec(name, schema, strict) — strict was serialized alongside json_schema instead of inside it. Old constructor preserved as @Deprecated for backwards compatibility; let me know if you prefer a clean breaking change instead.

vLLM vendor extensions via composable Filters (VllmFilters):
- ChatTemplateKwargs: injects chat_template_kwargs into requests
- ReasoningContent: maps reasoning_content to reasoning in responses
This is useful when for example using Qwen and you want to toggle thinking on/off. 

Fake generates synthetic logprobs when logprobs=true (streaming and non-streaming).

Unit tests passing, and verified against my local vLLM with Qwen3 / Phi.


connect/ai/openai/client/src/main/kotlin/org/http4k/connect/openai/OpenAI.kt
daviddenton, line 1: These extra functions should not appear in the adapter interface. The entire point of the connect pattern is that there is a clean separation between the adapter and the actions.

What are they here for? 🙃

connect/ai/openai/client/src/main/kotlin/org/http4k/connect/openai/VllmFilters.kt
daviddenton, line 1: Not sure if this belongs in this module since it's specific to a non-official OpenAI version of the API. In general we'd do this using a different module which is dedicated to that particular version of the API. I'm not too familiar with vLLM I'm afraid.

daviddenton:
Also - please ignore the Security CodeQL job - it doesn't support the latest version of Kotlin 🙄

connect/ai/openai/client/src/main/kotlin/org/http4k/connect/openai/VllmFilters.kt
torfinnberset: Indeed I wasn't sure either, I'll remove it for now, and if I see that there's more custom filters or modules needed for vllm, I can submit that separately

connect/ai/openai/client/src/main/kotlin/org/http4k/connect/openai/OpenAI.kt
torfinnberset: Sorry, I swore I saw this pattern elsewhere and I was just trying to be consistent 😅 I'll drop them

torfinnberset:
I've addressed your comments, thank you for your time!

connect/ai/openai/fake/src/test/kotlin/org/http4k/connect/openai/ChatCompletionJsonTest.kt
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+        val jsonSchema = parsed.response_format as JsonSchema
+        assertThat(jsonSchema.name, equalTo("response"))
+        assertThat(jsonSchema.strict, equalTo(true))
+        assertThat(jsonSchema.json_schema["type"], equalTo("object" as Any))
+    }
+
+    @Test
+    fun `legacy ChatCompletion JVM constructor is preserved`() {
```
daviddenton, line 155: we should call the old constructor if we want it to exist, :)

PR #1520, opened 2026-04-20
Add support for configuring ReturnResponse chaos behaviour remotely

jamieredding:
I'm unclear if this was omitted deliberately, as there was previously a disabled test for this when it was originally implemented.

Couple of notes:
1. I've not set a `x-http4k-chaos` header as the purpose of this is unclear to me. Happy for it to be included but possibly we don't want to just `toString()` the whole `ReturnResponse` into it?
   - I didn't like the idea of only adding the header when deserialising json, and not adding it when the response is provided via kotlin. But also didn't want to overwrite the user requested headers.
3. I've mirrored the json structure to be similar to the `ChaosTriggers.MatchRequest`. This means also not supporting multi-value headers. I don't have a use case for this and likely the trigger format would need to be revisited if we wanted support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ changes with their rationale when appropriate:
 
 ### v6.43.0.1 (uncut)
 - **http4k-ai-mcp-sdk**: [Fix] X402 and MPP payment filters fail with correct error message.
+- **http4k-testing-chaos**: Add `ReturnResponse` behaviour. H/T @jamieredding.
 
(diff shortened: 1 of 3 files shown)


daviddenton:
thanks for the contrib. The point of the chaos header is a marker that the response was generated by the ChaosEngine, so it should be present when the chaos is active. We don't need to include the entire thing - just so you can identify which type of Chaos was used.

daviddenton:

PR #1537, opened 2026-05-11
Move MimeTypes mapping file to standard resources directory to ease obfuscation

oharaandrew314:
Proguard is unable to adapt the resource path when it's in `META-INF`.  

The `loadMetaResource` utility function is internal, isn't used elsewhere, and isn't proguard friendly, so I removed it.

Afaik, shadow shouldn't require the `mergeServiceFiles` directive in order to migrate standard resources, so I removed the warning message.


oharaandrew314:
You know what.... it isn't all that useful to obfuscate library code, and I shouldn't be trying to do it.  So feel free to ignore this PR.

daviddenton:
TBH, I'm all for tidying up, but I'm a little worried about the blast radius, so I'm gonna take you advice 🙃

PR #1538, opened 2026-05-12
Fix for #1533: Aws sig4 path canonicalization needs double encoding 

### LINKED ISSUE #1533: AwsAuth: sigv4 path canonicalization regression for non-S3 services (since v6.5.3.0)
## Summary

  `org.http4k.filter.AwsAuth` produces sigv4 signatures that real AWS rejects with `InvalidSignatureException` for any URL path containing RFC 3986 reserved characters (verified: `:`). The regression was
  introduced in **v6.5.3.0** and is still present in the current latest **v6.46.1.0**.

  The bug appears to come from PR #1329 (*"Added support for s3 bucket keys containing hash (#)"* by @aharin), which modified the shared canonical-request code path while targeting an S3-specific change.

  ## Versions
  
  - **Last working**: `6.5.2.0`
  - **First failing**: `6.5.3.0`
  - **Still failing**: `6.46.1.0` (latest at time of report)

  ## Reproduction

  In our setup we sign requests via `AwsAuth` to a Lambda Function URL (service `lambda`, region `eu-north-1`). The differentiator between passing and failing requests is the URL path:

  | Request | Result |
  |---|---|
  | `GET /prompts` | ✅ AWS accepts the signature |
  | `GET /tasks/some-namespace:<UUID>/prompt` | ❌ `403 InvalidSignatureException` |
  | `POST /tasks` (no reserved chars in path) | ✅ |
  | `POST /tasks/some-namespace:<UUID>` | ❌ `403 InvalidSignatureException` |

  AWS response on failure:

  HTTP/1.1 403
  x-amzn-errortype: InvalidSignatureException
  {"message":"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method."}

  The differentiator across all our test cases is the presence of a colon (`:`) in a path segment. Paths free of reserved characters sign correctly; paths containing `:` produce signatures real AWS rejects.

  A minimal standalone reproducer comparing `AwsCanonicalRequest` output between v6.5.2.0 and v6.5.3.0 for the same input would be the cleanest artifact. I can put one together if it would help — wanted to
  file the bisection result first.

  ## Bisection

  6 rounds of binary bisection across the `6.5.2.0..6.46.1.0` range, running our internal e2e test that signs requests via `AwsAuth` against a real AWS Lambda Function URL:

  | Version | Result |
  |---|---|
  | 6.5.2.0 (last good) | ✅ |
  | **6.5.3.0** | **❌ first failing** |
  | 6.5.4.0 | ❌ |
  | 6.7.0.0 | ❌ |
  | 6.11.1.0 | ❌ |
  | 6.15.0.0 | ❌ |
  | 6.25.1.0 | ❌ |
  | 6.44.0.0 | ❌ |
  | 6.46.1.0 | ❌ |

  `6.5.2.1` does not exist on Maven Central or our private mirror, so the regressing change must be in the v6.5.2.0 → v6.5.3.0 commit range.

  ## Probable culprit

  PR #1329 (*"Added support for s3 bucket keys containing hash (#)"* by @aharin, merged 2025-04-01) is the only AWS-related change in that range per the v6.5.3.0 changelog entry. While the PR title is
  scoped to S3, the diff modifies the shared canonical-request code path:

  - `core/platform/aws/src/main/kotlin/org/http4k/aws/AwsCanonicalRequest.kt`
  - `core/platform/aws/src/main/kotlin/org/http4k/aws/AwsRequestPreSigner.kt`
  - `core/platform/aws/src/main/kotlin/org/http4k/aws/requestExtensions.kt`
  - `core/platform/aws/src/main/kotlin/org/http4k/filter/awsExtensions.kt`

  `AwsCanonicalRequest` is reached by every AWS service signing through `AwsAuth`, not only S3. The PR description also mentions a side-fix to `JavaHttpClient` for keys with spaces — suggesting the encoding
   logic was tightened in a way that helped some characters and broke others.

  ## Expected behavior

  Per [AWS sigv4 canonical request rules](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html):

  - **Non-S3 services**: path components are URL-encoded twice when constructing the canonical URI (each path segment URL-encoded, then the URI as a whole encoded again).
  - **S3**: path components are URL-encoded only once (S3 documents this as an exception).

  The pre-v6.5.3.0 behavior matched these rules and AWS accepted the signatures. Post-v6.5.3.0, paths containing reserved characters like `:` get a different (apparently single-encoded or
  differently-encoded) canonical form for non-S3 services, which AWS rejects.

  ## Impact

  This blocks every consumer of `AwsAuth` (i.e., every http4k user calling non-S3 AWS endpoints with paths that may contain reserved characters — UUIDs are fine, but namespaced identifiers like
  `tenant:resource-id`, ARNs in path, etc., are not). Pinning at v6.5.2.0 is the only viable workaround we've found, which blocks pulling in CVE fixes that ride on newer http4k transitives (Undertow, Netty,
   Jackson).
  
  ## Internal tracker

  We're tracking this in our private Jira as DPU-3046 (link won't be useful externally, just for our internal cross-reference).

  cc @aharin (PR #1329 author), @daviddenton (PR #1329 reviewer).

aharin:
This PR should fix #1533. For the purpose of sig4 generation, paths should be double url encoded fow all services except s3

### DIFF
--- a/core/platform/aws/src/main/kotlin/org/http4k/aws/AwsRequestPreSigner.kt
+++ b/core/platform/aws/src/main/kotlin/org/http4k/aws/AwsRequestPreSigner.kt
@@ -44,7 +44,7 @@ fun AwsRequestPreSigner(
         }
 
     val canonicalRequest =
-        AwsCanonicalRequest.of(fullRequest, payloadMode(request))
+        AwsCanonicalRequest.of(fullRequest.encodeUri(scope), payloadMode(request))
     val signature = AwsSignatureV4Signer.sign(canonicalRequest, scope, credentials, awsDate)
 
     AwsPreSignedRequest(
--- a/core/platform/aws/src/main/kotlin/org/http4k/aws/requestExtensions.kt
+++ b/core/platform/aws/src/main/kotlin/org/http4k/aws/requestExtensions.kt
@@ -8,6 +8,9 @@ import org.http4k.urlEncoded
 internal fun Request.encodeUri() =
     uri(uri.encodePathAndFragment())
 
+internal fun Request.encodeUri(scope: AwsCredentialScope) =
+    if (scope.service == "s3") this else encodeUri()
+
 private fun Uri.encodePathAndFragment() = if (fragment.isBlank())
     path(path.urlEncodedPath())
 else
--- a/core/platform/aws/src/main/kotlin/org/http4k/filter/awsExtensions.kt
+++ b/core/platform/aws/src/main/kotlin/org/http4k/filter/awsExtensions.kt
@@ -62,7 +62,7 @@ fun ClientFilters.AwsAuth(
                     } ?: this
                 }
 
-            val canonicalRequest = AwsCanonicalRequest.of(fullRequest, payload)
+            val canonicalRequest = AwsCanonicalRequest.of(fullRequest.encodeUri(scope), payload)
 
             val signedRequest = fullRequest
                 .replaceHeader("Authorization", buildAuthHeader(scope, credentials, canonicalRequest, date))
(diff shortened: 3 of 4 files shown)


daviddenton:
Nice, thanks! @aharin 🙃

daviddenton:

PR #1540, opened 2026-05-15
Add response expiration mechanism to cache filters

ollieabbey:
Unsure whether you think this a bit too opinionated for this library, but it feels like it's borderline and it's something that could be useful to others. If it's not something you think would be useful or fits then feel free to close. 

Basically this allows you to have a ReadWriteCache Filter that respects the cache control directives of a response, so consumers don't have to write logic to check this themselves. 

It uses some custom headers to store the expiry times alongside the response, which seems to be a pattern that's been used elsewhere (e.g. [here](https://github.com/http4k/http4k/blob/037f430dd7c5d699da7cdc7b2660b5c0ec6969ab/pro/wiretap/src/main/kotlin/org/http4k/wiretap/domain/TransactionMapping.kt#L33-L35)). I couldn't think of another simple way to store this information, but if there is then feel free to update or let me know what changes to make. These aren't surfaced to the caller outside of these filters.

By default this behaviour is disabled by passing a null param set, so consumers will have to opt in to get this behaviour and there should be no breaking changes.

### DIFF
--- a/core/tools/traffic-capture/src/main/kotlin/org/http4k/filter/TrafficFilters.kt
+++ b/core/tools/traffic-capture/src/main/kotlin/org/http4k/filter/TrafficFilters.kt
@@ -5,32 +5,151 @@ import org.http4k.core.Filter
 import org.http4k.core.Request
 import org.http4k.core.Response
 import org.http4k.core.Status
+import org.http4k.core.maxAge
+import org.http4k.core.staleIfError
+import org.http4k.core.staleWhileRevalidate
+import org.http4k.core.then
+import org.http4k.traffic.ReadWriteCache
 import org.http4k.traffic.Replay
 import org.http4k.traffic.Sink
 import org.http4k.traffic.Source
+import kotlin.time.Clock
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.Instant
 
 object TrafficFilters {
 
     /**
+     * Internal header names to store the expiration times of the response.
+     * Removed from the response and not returned to the caller, so ensure these don't clash with real headers.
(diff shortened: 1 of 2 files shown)


daviddenton:
Thanks for the contribution. We can see where it would be useful, but we're not sure that the current implementation is really workable as is. We passed it through Claude for a quick check and it came up with these points, which seem to all make sense 🙃 : 

  1. Default thread launcher is Thread { it.start() }. Unbounded thread creation per stale hit. Under any meaningful concurrency this is a footgun — at   least default to a single-thread executor, or document the risk.
  2. No single-flight on revalidation. N concurrent stale requests trigger N upstream calls. The whole point of stale-while-revalidate is to amortize  that.
  3. Internal state stored as headers on the cached Response (x-http4k-expiry, etc). The PR strips them on read, but anyone reading directly from the Source (bypassing the filter) sees them. Cache poisoning if an upstream legitimately sends those header names. Worth namespacing or moving the  metadata out-of-band.
  4. Sink/Source thread-safety isn't required by the existing contract, but background revalidation now assumes it. ReadWriteCache.Memory() happens to  be fine; custom implementations may not be.
  5. Cache-Control parsing is hand-rolled. Edge cases: quoted values, multiple directives, casing, no-store interaction. Worth verifying against  RFC 7234.
  6. Adds kotlinx.datetime.Instant/Clock to a module that previously didn't use it — check the module's existing deps. If new, that's a transitive dep  change for users.

ollieabbey:
Yeah all good points - will take a look when I get a chance and reopen.

daviddenton:

PR #1543, opened 2026-05-20
Fix Aws Auth content streaming + fix OKHttp client streaming

asadmanji:
I'm using the S3 PutObject example on https://www.http4k.org/ecosystem/connect/reference/amazon/s3/ to stream a large 1GB to S3 with OKHttp client. 
When trying to stream a large content payload to S3 I hit an out of memory exception because both the AwsAuth client filter and OKHttp client realise the body input stream. This PR attempts to fix both, and I've extended the example to include unsigned payloads as well.


daviddenton:

PR #1546, opened 2026-05-26
fix: handle brace quantifiers in UriTemplate regex patterns (#676)

### LINKED ISSUE #676: Regex using greedy qualifier for N times is quoted incorrectly

lukman48:
Fixes #676

## Problem
The `URI_TEMPLATE_FORMAT` regex `\{([^}]+?)(?::([^}]+))?\}` uses `[^}]+` for the regex part of path parameters. This fails when the user's regex contains `{n}` style quantifiers (e.g. `[a-z]{3}`) because the first `}` from the quantifier is matched as the closing brace of the template parameter.

For example, `/foo/{bar:[a-z]{3}}` would incorrectly parse the regex as `[a-z]{3}` and leave a trailing `}` unmatched, causing a `PatternSyntaxException`.

## Solution
Updated `URI_TEMPLATE_FORMAT` to properly handle balanced braces in the regex part by matching `{...}` groups:

```
\{([^:}]+)(?::((?:[^{}]|\{[^{}]*\})+))?\}
```

This pattern matches `{n}`, `{n,m}`, and other brace-containing regex constructs within the parameter regex while correctly finding the actual closing `}` of the template parameter.

## Changes
- `UriTemplate.kt`: Fixed `URI_TEMPLATE_FORMAT` regex to handle brace quantifiers
- `UriTemplateTest.kt`: Enabled previously disabled `greedyQualifiersAreNotReplaced` test and added tests for complex UUID regex patterns with multiple brace quantifiers and alternation

### DIFF
--- a/core/core/src/main/kotlin/org/http4k/core/UriTemplate.kt
+++ b/core/core/src/main/kotlin/org/http4k/core/UriTemplate.kt
@@ -17,7 +17,7 @@ data class UriTemplate private constructor(private val template: String) {
     private val parameterNames = matches.map { it.groupValues[1] }.toList()
 
     companion object {
-        private val URI_TEMPLATE_FORMAT = "\\{([^}]+?)(?::([^}]+))?\\}".toRegex() // ignore redundant warning #100
+        private val URI_TEMPLATE_FORMAT = "\\{([^:}]+)(?::((?:[^{}]|\\{[^{}]*\\})+))?\\}".toRegex() // ignore redundant warning #100
         fun from(template: String) = UriTemplate(template.trimSlashes())
 
         private fun String.trimSlashes() = "^(/+)?(.*?)(/+)?$".toRegex().replace(this) { result -> result.groupValues[2] }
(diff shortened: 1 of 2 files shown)


daviddenton:

PR #1567, opened 2026-06-26
fix incorrect gzip content length in streaming mode and add test

### LINKED ISSUE #1566: ResponseFilters.GZip with Streaming mode forwards stale Content-Length from upstream incorrectly
When `ResponseFilters.GZip(GzipCompressionMode.Streaming())` is applied to a response that already carries a Content-Length header (e.g. proxied from an upstream server), the header is forwarded unchanged even though the body has been replaced with a streaming gzip compressor of unknown length.

`CompressionResult.apply()` only updates `Content-Length` when `body.length != null` as such:

```kotlin
when {
    it.header("content-length") != null && body.length != null ->
        it.replaceHeader("content-length", body.length.toString())
    else -> it   // stale Content-Length passes through
}
```

`Body(GZippingInputStream(...))` always has `length == null`, so the else branch is always taken.

I noticed this because we serve .xlsx files through or server, proxying them from an upstream 'microservice', and these .xlsx files are already gzipped, in which case the gzip filter simply adds some header/trailer overhead but otherwise doesn't change the `Content-Length`. This resulted in the browser hanging forever when reading/inflating the response, because the gzip data was truncated due to a too-short content-length. In most cases, the content-length will be *larger* than the actual data, in which case we haven't seen this issue.

When Streaming mode is used and the body length is unknown, Content-Length should be removed from the response, e.g.

```kotlin
it.header("content-length") != null && body.length != null ->
    it.replaceHeader("content-length", body.length.toString())
it.header("content-length") != null && body.length == null ->
    it.removeHeader("content-length")   // length unknown after streaming gzip
else -> it
```

Incidentally, the last person to touch this code was my colleague @bjornbugge, but he didn't touch it enough it seems :D

adamschoenemann:
Fixes #1566

### DIFF
--- a/core/core/src/main/kotlin/org/http4k/filter/Gzip.kt
+++ b/core/core/src/main/kotlin/org/http4k/filter/Gzip.kt
@@ -63,6 +63,8 @@ data class CompressionResult(
                 when {
                     it.header("content-length") != null && body.length != null ->
                         it.replaceHeader("content-length", body.length.toString())
+                    it.header("content-length") != null && body.length == null ->
+                        it.removeHeader("content-length")
                     else -> it
                 }
             }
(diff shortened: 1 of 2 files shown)


daviddenton:

PR #1572, opened 2026-07-02
Fix issue posting a multipart form with no file selected in WebDriver implementation

akshay-agrawal04:
Hi!

Needed this fix for something at work. The issue is that when we post a multipart form without uploading a file, the WebDriver implementation returns a FileNotFoundException, which shouldn't be the case.

The test shows a reasonable example of "sending" a request without a file through a submit in the WebDriver, and we just assert that the request goes through when there are no files uploaded.

Thanks!


daviddenton:

PR #1583, opened 2026-07-09
add support for a JSON schema dialect selector in swagger ui

dzappold:


dzappold:
Can't reproduce that error locally.

PR #1585, opened 2026-07-09
MimeTypes to support files paths without extensions

oharaandrew314:
`MimeTypes.forFile` is unable to parse the extension for an extension-less file, e.g. `NOTICE`.  This prevents the `static` routing handler from serving extensionless-files, like `LICENSE`.  This is problematic when you're trying to do something like serving a generated license report.

This update will cause `MimeTypes` to return an empty-string for an unknown extension, allowing the content-type to be overridden generically.


oharaandrew314:
The failing test is just the flakey one disabled in the other PR :smile:

oharaandrew314:
I know this class is critical, so I've merged master back in to rerun the build.

PR #1588, opened 2026-07-17
DynamoDB: support ReturnValuesOnConditionCheckFailure on PutItem/UpdateItem/DeleteItem

torfinnberset:
DynamoDB can return the record that blocked a failed conditional write, but the top-level actions had no way to ask for it, so optimistic-lock callers pay a reload read per conflict just to see what they hit. 

The `ReturnValuesOnConditionCheckFailure` enum already existed and `TransactWriteItem` already used it; this brings `PutItem`/`UpdateItem`/`DeleteItem` in line.


daviddenton:
👍

PR #1599, opened 2026-08-05
Parse ConditionalCheckFailed from both DynamoDB and DynamoDB Local

torfinnberset:
DynamoDB names the error text `message`; DynamoDB Local names it `Message`. The AWS SDKs accept either, so reads here do too — otherwise a failed conditional write cannot be parsed at all against one of the two. Writes always use DynamoDB's own spelling.

The shared PutItem contract now asserts through the model rather than raw JSON, since both services run it.


daviddenton:
need to run spotlessApply to make the linter happy :)

torfinnberset:
Done 😅 Thanks!

daviddenton (approved):

Issue #1609, opened 2026-08-26
KotlinxSerialization.parse rejects top-level arrays, silently dropping OpenAPI examples for list bodies


cies:
### What happens

With `KotlinxSerialization` as the `Json` of an `OpenApi3` renderer, any operation whose body is a
top-level JSON array renders **without** its `example`. Object-bodied operations are fine, so a
document ends up inconsistent:

```
GET  /things       200  keys=['schema']              <-- List<Thing>, example dropped
POST /things       200  keys=['example', 'schema']   <-- Thing
```

Swapping the same contract to `Jackson` renders the example for both.

### Why

`ConfigurableKotlinxSerialization` parses with an *object* serializer:

```kotlin
// core/format/kotlinx-serialization/.../ConfigurableKotlinxSerialization.kt
override fun String.asJsonObject() = json.decodeFromString(JsonObject.serializer(), this)
```

`Json.parse` delegates to it (`core/format/core/.../Json.kt`):

```kotlin
fun parse(input: String): NODE = input.asJsonObject()
```

so parsing an array throws. `OpenApi3` takes a media type's example from the rendered message body
and [swallows the failure (for which I raised an issue of it's own)](https://github.com/http4k/http4k/issues/1608):

```kotlin
// core/api/openapi/.../OpenApi3.kt
SchemaContent(schema, message.bodyString().safeParse())

private fun String.safeParse(): NODE? = try { json.parse(this) } catch (e: Exception) { null }
```

`null` example → the key is omitted. Nothing logs, nothing fails.

### Reproduce

At the `Json` level:

```kotlin
KotlinxSerialization.parse("""[{"a":1}]""")
// Raises `JsonDecodingException: Expected class JsonObject ... but had JsonArray`
```

End to end:

```kotlin
@Serializable data class Thing(val name: String)

private val things = Body.auto<List<Thing>>().toLens()

val api = contract {
  renderer = OpenApi3(ApiInfo("example", "1.0"), json = KotlinxSerialization)
  descriptionPath = "/openapi.json"
  routes += "/things" meta {
    returning(OK, things to listOf(Thing("a")))
  } bindContract GET to { Response(OK).with(things of listOf(Thing("a"))) }
}

// paths./things.get.responses.200.content."application/json" has "schema" but no "example"
```

### Why I think it is a bug rather than a design choice

`kotlinx-serialization` is the only adapter that rejects arrays here — and Gson's message spells out
the intended contract:

[...]

PR #1610, opened 2026-08-26
Fix #1609

cies:
Fix for issue https://github.com/http4k/http4k/issues/1609

### DIFF
--- a/core/format/kotlinx-serialization/src/main/kotlin/org/http4k/format/ConfigurableKotlinxSerialization.kt
+++ b/core/format/kotlinx-serialization/src/main/kotlin/org/http4k/format/ConfigurableKotlinxSerialization.kt
@@ -79,15 +79,6 @@ open class ConfigurableKotlinxSerialization(
 
     override fun JsonElement.asCompactJsonString() = json.encodeToString(JsonElement.serializer(), this)
 
-    /**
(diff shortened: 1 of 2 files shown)


core/format/kotlinx-serialization/src/main/kotlin/org/http4k/format/ConfigurableKotlinxSerialization.kt
```diff
@@ -60,25 +60,35 @@ open class ConfigurableKotlinxSerialization(
...
         is JsonObject -> JsonType.Object
     }
 
     override fun JsonElement.asPrettyJsonString() = prettyJson.encodeToString(JsonElement.serializer(), this)
 
     override fun JsonElement.asCompactJsonString() = json.encodeToString(JsonElement.serializer(), this)
 
-    override fun String.asJsonObject() = json.decodeFromString(JsonObject.serializer(), this)
+    /**
+     * Objects and arrays both parse, as they do in every other [Json] implementation: the name is
```
cies: maybe the comment here is too much.

i can reduce it (remove the reference to how things were) or remove it all together.

daviddenton: yeah - if you could remove the comment and then we'll merge this :)

